PR #189, opened 2017-04-19
Now pass pool_size to the underlayed aiomcache

abusi:
We needed to tweak this on our side. cause the default for aiomcache.client is 2, and it is really light in a context of high polling a memcache.


argaen:
Thanks for your contribution!

PR #200, opened 2017-04-23
Remove unnecessary warning

pitcons:
There is no cPickle in python 3


argaen:
You're right, thanks!

PR #213, opened 2017-05-06
Decorators support introspection

### LINKED ISSUE #212: Use decorator module for signature-preserving decorators
It looks like your awesome `cached` and `multi_cached` decorators don't make use of https://pypi.python.org/pypi/decorator which helpers write cleaner decorators that benefit from signature-preservation (which is important to some people :-)

I could submit a PR if you like?

argaen:
Closes #212

### DIFF
--- a/tests/ut/test_decorators.py
+++ b/tests/ut/test_decorators.py
@@ -1,6 +1,7 @@
 import sys
 import pytest
 import random
+import inspect
 import asynctest
 
 from unittest import mock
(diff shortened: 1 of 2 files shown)


tests/ut/test_decorators.py
```diff
@@ -197,6 +198,16 @@ class Dummy:
...
 
+    @pytest.mark.asyncio
+    async def test_cached_keeps_signature(self):
+        @cached()
+        async def what(self, a, b):
+            return "1"
+
+        assert what.__name__ == "what"
+        assert str(inspect.signature(what)) == '(self, a, b)'
+        assert inspect.getfullargspec(what.__wrapped__).args == ['self', 'a', 'b']
```
argaen: @OOPman have a look at the behavior here. Dunno how zope does the interface check, but it should use the `signature` (https://docs.python.org/3/library/inspect.html#inspect.signature) one since it enables the `follow_wrapper` by default.

getfullargspec (https://docs.python.org/3/library/inspect.html#inspect.getfullargspec) doesn't support following the __wrapped__ attribute that `functools.wraps` adds.

OOPMan: Looks good. Iirc Zope uses inspect as you have since it's the only sane way to do it. There are probably some insane ways but why worry about those :-)

argaen: yeah, my point was more that if Zope uses `inspect.signature` then it will work. If it uses `inspect.getfullargspec` it won't because the real signature is under the `__wrapped__` attribute of the function and `getfullargspec` doesn't use it (while `signature` does :))

Issue #27, opened 2016-10-23
Implement locking to avoid dogpile effects


argaen:
This would help to avoid problems like race conditions, dogpile effect, etc.

Some ideas:
- https://bitbucket.org/zzzeek/dogpile.cache/
- https://pypi.python.org/pypi/HermesCache

References:
- http://martin.kleppmann.com/2016/02/08/how-to-do-distributed-locking.html
- http://antirez.com/news/101
- https://aphyr.com/posts/283-jepsen-redis

Couple of things to keep in mind:

- Keep this as decoupled as possible, I don't know yet all use cases but for consistency, the approaches described below are not the most recommended. Since the main use case now is avoid stampede and avoid race conditions, we are "fine".
- Should add disclaimers that the lock doesn't aim for consistency.

**Redis**

For now I'm thinking about going for using the simple approach described in https://redis.io/topics/distlock#correct-implementation-with-a-single-instance (lease time should be able to be passed by the user).

**Memcached**

https://bluxte.net/musings/2009/10/28/simple-distributed-lock-memcached/
https://github.com/memcached/memcached/wiki/ProgrammingTricks#avoiding-stampeding-herd

**Memory**

Just go with asyncio.Lock

argaen:
This is not a policy and should only be implemented in the decorators (which are the ones surrounding the functions). Explicit usages are not meant for this.

argaen:
Closed with 

#235 Redis redlock
#240 Memcached redlock
#241 Memory redlock

PR #235, opened 2017-05-23
Feature #27/redis locking

argaen:
Adds support for Redis distributed lock to avoid cache stampede effects. For now its not applied anywhere and functions/class is private.

Once implementation for memcached and memory is done, I'll add support for cached decorator at first and make it public as a on demand feature after. The implementation for memcached and memory won't be hard now that the base is set with Redis.

Some things to still add in this MR:

- ~Lease time should support miliseconds (now its just seconds). I think this will raise a problem for memcached since it doesn't support it?~
- There is a worst case scenario for operations that take the whole lease time N. If a call arrives 0.1 ms before the lease expires, the call will be blocked N while it would only need 0.1 to continue. Gonna leave it like that (documented in the class).
- If the lease expires, ALL calls will try to retrieve the value from the resource that its being protected. I've decided to do it this way to avoid block the calls too much waiting for leases that could always be expiring. Gonna leave it like that (documented in the class).

@gjcarneiro what do you think, is there anything you are missing? You can see an example of it working in https://github.com/argaen/aiocache/compare/feature_%2327/redis_locking?expand=1#diff-61192af138ef98c9b82cba13168e6e12R205 (the `asyncio.sleep` there represents the expensive operation)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -91,3 +91,5 @@ ENV/
 
 dump.rdb
 _release_notes
+
+tags
--- a/aiocache/_lock.py
+++ b/aiocache/_lock.py
@@ -2,11 +2,19 @@ import asyncio
 import uuid
 
 
-class _Lock:
+class _DistributedLock:
     """
     Implementation of [Redlock](https://redis.io/topics/distlock)
     with a single instance because aiocache is focused on single
     instance cache.
+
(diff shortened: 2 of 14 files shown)


argaen:
Redis integration done. #27

Issue #233, opened 2017-05-16
Resource Cleanup


Quinny:
Maybe I am using it wrong, but when I tag functions with the `cached` decorator for a redis cache like so:

<pre>
@cached(ttl=ONE_HOUR, cache=RedisCache, key="rssfeed",
        serializer=PickleSerializer(), port=6379, namespace="main",
        endpoint="192.168.1.19")
async def f(....):
  .....
</pre>

I get the following errors upon exit:

<pre>
2017-05-16 11:32:20,999 ERROR asyncio(0) | Task was destroyed but it is pending!
task: <Task pending coro=<RedisConnection._read_data() done, defined at /usr/local/lib/python3.6/site-packages/aioredis/connection.py:127> wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x102d290a8>()]> cb=[Future.set_result()]>
</pre>

It seems that the redis connections are not being properly handled on exit.  Is there currently a way to deal with this?

I am running a sanic app with python3.6 if that makes a difference.

Thanks! Awesome project by the way.

argaen:
When you say "It seems that the redis connections are not being properly handled on exit" you mean when exiting your program (hitting Ctrl-C or because your code crashes) right?

If that's the case yeah its normal, I haven't exposed any option to clean up resources once the program ends with aioredis (aiomcache doesn't support that). I'll leave this open because its something that's been annoying me for a while but was working other things I considered more prioritary.

Out of curiosity, is it causing any issues to you?

Quinny:
Yes, I mean hitting Ctrl-C to kill the sanic web server.  It doesn't seem to be causing any issues, it just kinda bothers me to see a spew of error messages every time I bring down my server.

One suggestion I have that may help with this: Instead of the cache decorator taking a class, it should take an instance of a cache, and use that instance instead of constructing one.  Then, that instance can be properly closed upon program exit.  For example:

<pre>
redis_cache = RedisCache(host=....)

@cached(cache=redis_cache)
async def f(...):
  ....

if __name__ == "__main__":
  ....
  await redis_cache.close()
</pre>

argaen:
Nope, I don't like your proposal for 2 reasons:

- Client will always have to initialize the cache object as a global variable. Decorators are evaluated at import time so this means the cache should be initialized also during import time always and thats a code smell for that use case.
- We would break the contract and I don't think it is worth it for the problem it is solving.

There are three situations I can think off that should be covered:

[...]

PR #236, opened 2017-05-23
#233 Added interface for performing resource clean up and updated tests

Quinny:


aiocache/backends/redis.py
```diff
@@ -121,6 +121,13 @@ def __init__(
             kwargs["encoding"] = encoding
         return await getattr(_conn, command)(*args, **kwargs)
 
+    @conn
+    async def _close(self, *args, _conn=None, **kwargs):
```
argaen: Close doesn't need the `_conn` to be propagated

aiocache/base.py
```diff
@@ -429,6 +429,25 @@ def plugins(self, value):
     async def _raw(self, command, *args, **kwargs):
         raise NotImplementedError()
 
+    @API.register
+    @API.aiocache_enabled()
+    @API.timeout
+    @API.plugins
```
argaen: Can you remove `register`, `aiocache_enabled` and `plugins` decorators? Those decorators are for cache specific commands

aiocache/decorators.py
```diff
@@ -55,7 +55,9 @@ def cached_decorator(func):
 
             try:
                 if await cache_instance.exists(cache_key):
-                    return await cache_instance.get(cache_key)
+                    result = await cache_instance.get(cache_key)
+                    await cache_instance.close()
```
argaen: Hmmmm we don't need to block the user until the pool is closed. `asyncio.ensure_future` would be sufficient here

aiocache/decorators.py
```diff
@@ -67,6 +69,7 @@ def cached_decorator(func):
             except Exception:
                 logger.exception("Unexpected error with %s", cache_instance)
 
+            await cache_instance.close()
```
argaen: Same

aiocache/decorators.py
```diff
@@ -145,6 +148,7 @@ def multi_cached_decorator(func):
                     except Exception:
                         logger.exception("Unexpected error with %s", cache_instance)
 
+            await cache_instance.close()
```
argaen: Same

examples/cached_alias_config.py
```diff
@@ -33,6 +33,7 @@
     assert await cache.get("key") == "value"
     assert isinstance(cache, SimpleMemoryCache)
     assert isinstance(cache.serializer, DefaultSerializer)
+    await cache.close()
```
argaen: This is not ideal. `caches.get('default'` returns a singleton identified by 'default'. User may want to use this cache in another function.

Global aliased caches should be closed when the program exits

argaen:
Thanks for your work, looks good! One extra comment, could you try to not decrease the coverage? :). You can check what you are missing with `make cov`

aiocache/decorators.py
Quinny: I agree, but won't that bring that the warnings? The future created must still be awaited at some point.

examples/cached_alias_config.py
Quinny: The user can choose to close this cache whenever they want.  This is just an example usage, no?

aiocache/backends/redis.py
Quinny: Done

aiocache/base.py
Quinny: Done.

Quinny:
All checks now passing

examples/cached_alias_config.py
argaen: yup, its just for the sake of "good" example. Its not ideal to close a singleton cache in the middle of a cache :)

aiocache/decorators.py
argaen: I agree with you on reusing the connection for the decorators, this is something I have in mind because there are a couple of edge cases.

Regardless of reusing or not the connection, its not worth it to delay the response to the user waiting for the close to pool. I prefer having a message saying Task X wasn't executed rather than unneeded delayed responses. We can think how to close those pending connections later :)

examples/cached_alias_config.py
Quinny: Fair enough, I've moved the closing of the cache into the body of test_alias

aiocache/decorators.py
Quinny: I see where you are coming from, but I don't think this "delay" is as big of a deal as you are making it out to be.  Since the code is using asyncio it is not as if the program will be idle whilst closing the connection, the event loop will context switch to another task.

I have made the changes and the error messages have now returned.  I would advise either reverting back to how I originally wrote it (awaiting the close), or reusing the same cache instance within a decorator.  Ultimately it is up to you.

argaen: Yeah, I'm not talking about this kind of speed but speed for a single call. In order for the function to return, it will have to wait for the pool to close (and the loop to reschedule it) when it really doesn't have to to retrieve the result.

In an API with high load, this can lead to unnecessary delayed calls. Imagine the following case:

1. New call arrives to the decorated function F
2. Other 1000 async calls arrive (no need to be on the same function, any async will do)
3. F gets the loop and executes until arriving to the `await close`. (I know there are more breaks in the middle but for the sake of simplicity lets imagine we are there already). Because of finding the `await` it is rescheduled to the queue of tasks to execute (as the last one because its round robin).
4. Until all 1000 async tasks aren't processed (i.e. find their first break), the loop won't pick again  F and it will pick it just to return something it could have returned already.

Also there can be side effects when trying to close the connection (https://github.com/aio-libs/aioredis/blob/master/aioredis/connection.py#L315) that raise exceptions.

Because all of that I prefer to use `ensure_future`. I agree with you that its a code smell and we won't be freeing all the resources but I prefer it that way. I have in mind a way to solve these situations, when I mature the idea a bit more I'll ping you to validate it if that's okay.

I hope this convinces you :)

Quinny: Not entirely, I guess we will agree to disagree haha.  I have made the changes to use ensure future anyways.

argaen:
Thanks for your contribution. As said I'll ping you when I have the solution for cleaning up the rest of resources :)

PR #261, opened 2017-06-08
Don't acquire and hold a connection in decorator argaen/aiocache#260

dmzkrsk:
argaen/aiocache#260

### DIFF
--- a/aiocache/decorators.py
+++ b/aiocache/decorators.py
@@ -46,16 +46,11 @@ class cached:
         self.noself = noself
         self.alias = alias
         self.cache = None
-        self._conn = None
 
         self._cache = cache
         self._serializer = serializer
         self._plugins = plugins
         self._kwargs = kwargs
-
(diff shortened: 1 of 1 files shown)


aiocache/decorators.py
```diff
@@ -72,18 +67,21 @@ def __call__(self, f):
...
 
     async def decorator(self, f, *args, **kwargs):
-        self._conn = self.cache.get_connection()
-
-        async with self._conn:
-            key = self.get_cache_key(f, args, kwargs)
+        key = self.get_cache_key(f, args, kwargs)
 
-            value = await self.get_from_cache(key)
+        try:
```
argaen: Please keep the `get_from_cache` and `set_in_cache` functions. In the end both `cached` and `cached_stampede` will use the same code now :)

argaen:
Closing this in favor of #267

PR #271, opened 2017-06-17
multi_cached can use arg for key_from_attr

### LINKED ISSUE #270: multi_cached function call requires parameter name
Basic case:
```python
@multi_cached('ids', ttl=10)
async def get_async(ids):
    return {1: '1', 2: '2'}

res = await get_async([1, 2])
```
TypeError: get_async() got multiple values for argument 'ids'
/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/aiocache/decorators.py:254: TypeError

With additional decorator, there is another error:
```python
@multi_cached('ids', ttl=10)
@stats.timer('time')
async def get_async(ids):
    return {1: '1', 2: '2'}

res = await get_async([1, 2])
```
KeyError: 'ids'
/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/aiocache/decorators.py:262: KeyError

Workaround:
```python
res = await get_async(ids=[1, 2])
```

argaen:
before only params defined in kwargs where working due to the
behavior defined in _get_args_dict function. This has now been
fixed and it behaves as expected.

Closes #270

### DIFF
--- a/aiocache/decorators.py
+++ b/aiocache/decorators.py
@@ -174,7 +174,7 @@ def _get_cache(
 def _get_args_dict(func, args, kwargs):
     defaults = {
         arg_name: arg.default for arg_name, arg in inspect.signature(func).parameters.items()
-        if arg.default is not inspect._empty
+        if arg.default is not inspect._empty   # TODO: bug prone..
     }
     args_names = func.__code__.co_varnames[:func.__code__.co_argcount]
     return {**defaults, **dict(zip(args_names, args)), **kwargs}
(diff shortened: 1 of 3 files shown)


argaen:
Acceptance added for this in 
- https://github.com/argaen/aiocache/pull/271/files#diff-060a5fa57cd39d163c070271447a598dR136
- https://github.com/argaen/aiocache/pull/271/files#diff-060a5fa57cd39d163c070271447a598dR146

PR #324, opened 2017-08-07
Add support for retrieving ttl of a key in redis

Jerevia:


argaen:
Hi @Jerevia.

Sorry but I can't accept this PR for one main reason: To expose a command in the public API, the functionality must be implemented in the three different backends (memory and memcached are missing right now).

Because of this, for a specific backend some functionality may be lost but that's why the `raw` command exists. To do `ttl` for redis you could do the following:

```
await cache.raw("ttl", "namespace:my_key")
```

I hope it helps

PR #344, opened 2017-10-12
aiocache compatible with aioredis 1.0.0b2

masipcat:
Self-explanatory


masipcat:
This PR adds compatibility with aioredis 1.0.0b2 and keeps backwards compatibility with aioredis v0.3.3

aiocache/backends/redis.py
```diff
@@ -8,11 +8,18 @@
 from aiocache.serializers import JsonSerializer
 
 
+is_aioredis_before_major_one = aioredis.__version__.startswith('0.')
```
argaen: Can you update the `setup.py` file to reflect the new supported version? Now the range is set to `'aioredis>=0.3.3,<1'`

argaen: Also, can you make the name less verbose? something like `aioredis_before_one`

masipcat: I think supporting version 1.0.0 beta 2 (`'aioredis>=0.3.3,<=1.0.0b2'`) by default isn't a good idea until is stable.

If someone wants to use aiocache with aioredis 1.0.0b2 should define explicitly:
```
aioredis==1.0.0b2
aiocache
```
instead of
```
aiocache[redis]
```

masipcat:
I don't know how to add missing tests

aiocache/backends/redis.py
argaen: Sounds good to me

argaen:
Will add them in this PR when I have time them. Thanks :)

PR #360, opened 2017-11-18
WIP Support aioredis==1.0

### LINKED ISSUE #345: aiocache compatible with aioredis 1.0.0b2
I created pull request #344 which implements support for aioredis 0.3.3 and 1.0.0b2.

argaen:
Closes #345 

Reduces coverage but because its not being combined, this will be fixed in #361

### DIFF
--- a/aiocache/backends/redis.py
+++ b/aiocache/backends/redis.py
@@ -15,11 +15,12 @@ def conn(func):
     @functools.wraps(func)
     async def wrapper(self, *args, _conn=None, **kwargs):
         if _conn is None:
-            if AIOREDIS_BEFORE_ONE:
-                with await self._connect() as _conn:
-                    return await func(self, *args, _conn=_conn, **kwargs)
-            else:
-                _conn = await self._connect()
+
+            pool = await self._get_pool()
+            conn_context = await pool
+            with conn_context as _conn:
+                if not AIOREDIS_BEFORE_ONE:
+                    _conn = aioredis.Redis(_conn)
                 return await func(self, *args, _conn=_conn, **kwargs)
 
         return await func(self, *args, _conn=_conn, **kwargs)
@@ -65,16 +66,17 @@ class RedisBackend:
         self._pool = None
 
     async def acquire_conn(self):
-        await self._connect()
-        if AIOREDIS_BEFORE_ONE:
-            return await self._pool.acquire()
-        return (await self._pool)
+        await self._get_pool()
+        conn = await self._pool.acquire()
+        if not AIOREDIS_BEFORE_ONE:
+            conn = aioredis.Redis(conn)
+        return conn
 
     async def release_conn(self, _conn):
         if AIOREDIS_BEFORE_ONE:
             self._pool.release(_conn)
         else:
-            self._pool._pool_or_conn.release(_conn._pool_or_conn)
+            self._pool.release(_conn.connection)
 
     @conn
     async def _get(self, key, encoding="utf-8", _conn=None):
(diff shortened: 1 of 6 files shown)


aiocache/backends/redis.py
```diff
@@ -65,12 +65,16 @@ def __init__(
...
+        if AIOREDIS_BEFORE_ONE:
+            return await self._pool.acquire()
+        return (await self._pool)
 
     async def release_conn(self, _conn):
-        self._pool.release(_conn)
+        if AIOREDIS_BEFORE_ONE:
+            self._pool.release(_conn)
+        else:
+            self._pool._pool_or_conn.release(_conn._pool_or_conn)
```
argaen: meh.. need to check if there is a better way to release a connection...

aiocache/backends/redis.py
```diff
@@ -185,7 +189,8 @@ def __init__(
 
     async def _close(self, *args, **kwargs):
         if self._pool is not None:
-            await self._pool.clear()
+            if AIOREDIS_BEFORE_ONE:
```
argaen: There is no clear anymore? :thinking:

tests/acceptance/test_base.py
```diff
@@ -267,4 +268,5 @@ class TestRedisCache:
     async def test_close(self, redis_cache):
         await redis_cache.set(pytest.KEY, "value")
         await redis_cache._close()
-        assert redis_cache._pool.size == 0
+        if AIOREDIS_BEFORE_ONE:
+            assert redis_cache._pool.size == 0
```
argaen: If clear is available somewhere, need to update here too so size is checked

tests/ut/backends/test_redis.py
```diff
@@ -6,7 +6,7 @@
 from aiocache import RedisCache
 from aiocache.base import BaseCache
 from aiocache.serializers import JsonSerializer
-from aiocache.backends.redis import RedisBackend, conn
+from aiocache.backends.redis import RedisBackend, conn, AIOREDIS_BEFORE_ONE
 
 
 class FakeRedisConn:
```
argaen: Need to revisit this and `FakePool`. The whole thing is starting to become a mess and hard to understand. Maybe its worth to just separate in 2 classes. One for each version of redis and then let the fixture decide when to load one or the other

tests/ut/backends/test_redis.py
```diff
@@ -345,7 +347,8 @@ def test_setup_override(self):
         cache._pool = pool
         await cache._raw("set", pytest.KEY, 1)
         await cache._close()
-        assert pool.clear.call_count == 1
+        if AIOREDIS_BEFORE_ONE:
```
argaen: same here for the clear thing

argaen:
Hey @popravich do you mind having a look at this PR? Updated aiocache to support new aioredis release but there are couple of things that maybe I'm doing wrong (see the comments)

popravich:
Hi @argaen ,
In your case, instead of using `create_redis_pool` and lots of `if AIOREDIS_BEFORE_ONE`
I would recommend using raw pool and `Redis` class so you will have less changes to your code,
here is the example of what I mean:
```python
# before v1.0
# self._pool = await create_redis(...)
redis = await self._pool.acquire()
self._pool.release(redis)

# since v1.0
# self._pool = await create_pool(...)
conn = await self._pool.acquire()
redis = Redis(conn)
self._pool.release(redis.connection)  # connection property returns _pool_or_conn attribute
```

argaen:
@popravich done the changes according to your recommendation: https://github.com/argaen/aiocache/pull/360/files#diff-0509a8110645553d3299c7430fd7c554

Let me know if that look reasonable and I'll merge :)

popravich:
Looks good

PR #368, opened 2017-12-22
Added create_connection_timeout for redis>=1.0.0 on creating connecti…

tmarques82:
…ons pool

### DIFF
--- a/aiocache/backends/redis.py
+++ b/aiocache/backends/redis.py
@@ -53,7 +53,7 @@ class RedisBackend:
 
     def __init__(
             self, endpoint="127.0.0.1", port=6379, db=0, password=None,
-            pool_min_size=1, pool_max_size=10, loop=None, **kwargs):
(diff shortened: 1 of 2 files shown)


tmarques82:
@argaen, codecov has reported a decrease on coverage due to a condition which is consider a partial hit. Concretely the line https://github.com/argaen/aiocache/pull/368/files#diff-0509a8110645553d3299c7430fd7c554R208.

What do you prefer to do there?
- Keep it as it is and decreasing the coverage
- Run a build step for codecov considering deps-lowest in order to have aioredis==0.3.3, so that line would give a negative result. That would also increase time of CI.

argaen:
Hmmm if I'm not mistaken the build with the deps-lowest should be run so coverage shouldn't be decreasing. There are other conditionals around with the redis version and was working fine.

Sadly I don't have the laptop around to be able to check it properly, so I won't be able to have a look until I get back :(

PR #382, opened 2018-03-02
upgrade aioredis to 1.1.0

jairhenrique:


argaen:
Hi @jairhenrique is there any reason why you're setting minimum version to 1.1.0? I'm -1 on being that breaking upgrading dependencies. I would start putting a warning if someone is using below 1.0 and in next release we could enforce `>1.0`

jairhenrique:
@argaen aioredis is a dependency of aiocache, in my opinion, if someone uses aiocache defining which version of aioredis they will use, it is wrong.

So, I do not see a problem in updating the dependency that should only be internal to aiocache.

argaen:
I don't agree, in general its good to leave dependencies open until a reasonable point. Some reasons:
- What if another package has also a dependency with aioredis? If you force strict dependencies you as a user have big changes of having version clashes between your libraries
- Version upgrades should be done smoothly so users can upgrade easily. I.e. having a release where you warn users versions <1 are not supported anymore and in the next release you enforce >1.

In general its a good practice to leave dependencies as open as possible and consider upgrading when:
1. Supporting all versions becomes painful because of many breaking changes from that dependency.
2. Doesn't make sense anymore to support older versions (I don't think a reasonable time has passed yet since 1.0 was released for aioredis)

Issue #385, opened 2018-03-12
Better way to add single entries to cache config


jcugat:
If I want to add a single new configuration cache without removing existing ones, I have to do some work in the `set_config()` call:

```py
caches.set_config({
    **caches.get_config(),
    'new_cache': {
        'cache': 'aiocache.SimpleMemoryCache',
        'namespace': 'new_namespace',
    },
})
```

What do you think about adding a new method, `add_config(name, <params>)` that just adds a new configuration cache?

```py
caches.add_config('new_cache', cache='aiocache.SimpleMemoryCache', namespace='new_namespace')
```

argaen:
The way the factory/cache manager was thought originally was a way to set up a set of caches that would then later be used in the different services.

In general the pattern is:
1. Have a config file defining the caches you want
2. In app startup time call caches.set_config() with the whole configuration
3. Use the aliases later in the code

What's your use case? Are you initializing each alias in the service constructors? Is there a specific reason why you doing it this way?

jcugat:
I have a project with several client libraries which are completely decoupled from the main app. It made sense to have their configuration separated, inside their own file, instead of the main app.

argaen:
Then those caches are local to the services? Why don't you create the cache explicitly or using `caches.create`? No need to call `caches.set_config` and retrieve them with an alias.

Naming caches with an alias and storing their config with `caches.set_config` is in case you want to reuse them in any part of your code.

PR #401, opened 2018-05-09
Factory refactoring

vovanbo:
- Use MappingProxyType to avoid deep copying of cache config in factory
- Use property getter and setter to handle config instead of two methods
- Add deprecation warnings for these methods
- Fix cache serializer and plugins initialisation to avoid deep copying
- Use import_module to simplify creation of cache class from string
- Fix docs formatting
- Update docs and tests
- Fix runners of flake8 and coverage in Makefile
- Expand extras for development requirements
- Upgrade async-timeout to 3.0.0

P.S. Excuse me for huge pull request :)

### DIFF
--- a/aiocache/factory.py
+++ b/aiocache/factory.py
@@ -154,6 +154,9 @@ class CacheHandler:
 
     def get_config(self):
         """
+        .. deprecated:: 0.9.1
+            Use ``config`` property getter instead.
+
         Return a proxy to current stored config
         """
         warnings.warn(
@@ -172,6 +175,9 @@ class CacheHandler:
 
     def set_config(self, config: MutableMapping):
         """
+        .. deprecated:: 0.9.1
+            Use ``config`` property setter instead.
+
         Set (override) the default config for cache aliases from a dict-like
         structure. The structure is the following::
 


Makefile
```diff
@@ -2,7 +2,7 @@ cov-report = true
 
 
 lint:
-	flake8
+	pipenv run flake8
```
argaen: This breaks the pipeline (https://travis-ci.org/argaen/aiocache/jobs/377003392#L483), you need to install pipenv in the syntax step too (tox file)

aiocache/factory.py
```diff
@@ -1,25 +1,31 @@
...
 def _create_cache(cache, serializer=None, plugins=None, **kwargs):
-
     if serializer is not None:
-        cls = serializer.pop("class")
+        cls = serializer['class']
         cls = _class_from_string(cls) if isinstance(cls, str) else cls
-        serializer = cls(**serializer)
+        serializer = cls(
+            **{k: v for k, v in serializer.items() if k != 'class'}
+        )
```
argaen: how is this better than before? Previous version looks simpler

aiocache/factory.py
```diff
@@ -1,25 +1,31 @@
...
     plugins_instances = []
     if plugins is not None:
         for plugin in plugins:
-            cls = plugin.pop("class")
+            cls = plugin['class']
             cls = _class_from_string(cls) if isinstance(cls, str) else cls
-            plugins_instances.append(cls(**plugin))
+            plugins_instances.append(
+                cls(**{k: v for k, v in plugin.items() if k != 'class'})
+            )
```
argaen: same as above

aiocache/factory.py
```diff
@@ -30,12 +36,14 @@ def _create_cache(cache, serializer=None, plugins=None, **kwargs):
...
+    """
+    Cache handler.
+    """
     _config = {
         'default': {
-            'cache': "aiocache.SimpleMemoryCache",
+            'cache': 'aiocache.SimpleMemoryCache',
             'serializer': {
-                'class': "aiocache.serializers.StringSerializer"
+                'class': 'aiocache.serializers.StringSerializer'
```
argaen: thanks :)

aiocache/factory.py
```diff
@@ -79,25 +88,98 @@ def create(self, alias=None, cache=None, **kwargs):
...
+            '"get_config" method is deprecated. Use "config" property instead.',
+            DeprecationWarning
+        )
+        return self.config
+
+    @config.setter
+    def config(self, value: MutableMapping):
+        if 'default' not in value:
+            raise ValueError('default config must be provided')
+        for config_name in value.keys():
```
argaen: no need for `.keys`. It already iterates over keys

aiocache/factory.py
```diff
@@ -79,25 +88,98 @@ def create(self, alias=None, cache=None, **kwargs):
         cache = _create_cache(**{**config, **kwargs})
         return cache
 
-    def get_alias_config(self, alias):
-        config = self.get_config()
-        if alias not in config:
+    def get_alias_config(self, alias) -> MappingProxyType:
+        if alias not in self._config:
```
argaen: minor: maybe better to use `self.config`

argaen:
@jcugat if you can have a look at the PR it would help, better to have another couple of eyes reviewing :)

jcugat:
Sure! I'll take a look this weekend.

aiocache/factory.py
```diff
@@ -79,25 +88,98 @@ def create(self, alias=None, cache=None, **kwargs):
...
         """
-        return deepcopy(self._config)
+        warnings.warn(
+            '"get_config" method is deprecated. Use "config" property instead.',
+            DeprecationWarning
+        )
+        return self.config
+
+    @config.setter
+    def config(self, value: MutableMapping):
```
jcugat: I understand that the main advantage of `MappingProxyType` over plain dicts is that we can avoid deepcopying it. But what about if a user provides a dict to the setter and then modifies that dict? It would cause the dict inside aiocache to be modified too right? This might introduce subtle bugs :/

aiocache/factory.py
jcugat: I guess we can't use `pop()` anymore since the dict is not copied, and we would be modifying it.

argaen:
> but unfortunately goes a bit against the issue I raised in #385 (I still have to think some alternative to it).
I think the comment I posted in that other issue still applies. We can discuss it there. Anyway yeah I like the changes here too

aiocache/factory.py
argaen: that's a valid point, maybe we can still have to use `deepcopy` here

aiocache/factory.py
vovanbo: It's just to avoid creation of MappingProxyType instance in property getter

Makefile
vovanbo: I'll fix it soon

aiocache/factory.py
vovanbo: Yes, exactly, it was changed to avoid `pop()` usage here

Issue #386, opened 2018-03-12
Disable decorator cache per-call


jcugat:
Sometimes we want to disable the cache for specific calls to a decorated method.

Adding a keyword argument (like `aiocache_disable=True`) that is consumed by the decorators and disables the cache for that specific call.

In this case, we don't try to load from the cache, but should we cache the result? My guess would be no, but interested in your thoughts around this.

argaen:
I think it's an interesting feature to have. We could have two different args:
- ignore_read
- ignore write

jcugat:
Yeah, that's way better. So the approach would be something like:

- `aiocache_disable_read`: does not try to retrieve the value from the cache, but stores the result there
- `aiocache_disable_write`: tries to read from cache first, if it's found returns it. In case it's not found, makes the original call, but does not save the result in the cache.

Then we wouldn't need the general `aiocache_disable` param, we could obtain the same behavior disabling both read and write.

argaen:
Closing, just merged #404. Thanks!

PR #404, opened 2018-05-26
Disable decorators cache per-call

jcugat:
MR for #386 

Not convinced about the naming (too long?), I'm open to suggestions.

Regarding the documentation, I am not sure where to put it since the docstring in the decorator is only for parameters for the decorator itself, and not calls to the decorator.

### DIFF
--- a/aiocache/decorators.py
+++ b/aiocache/decorators.py
@@ -72,14 +72,19 @@ class cached:
         return wrapper
 
     async def decorator(self, f, *args, **kwargs):
+        disable_read = kwargs.pop('aiocache_disable_read', False)
+        disable_write = kwargs.pop('aiocache_disable_write', False)
         key = self.get_cache_key(f, args, kwargs)
 
-        value = await self.get_from_cache(key)
-        if value is not None:
-            return value
+        if not disable_read:
+            value = await self.get_from_cache(key)
+            if value is not None:
+                return value
 
         result = await f(*args, **kwargs)
-        await self.set_in_cache(key, result)
+
+        if not disable_write:
+            await self.set_in_cache(key, result)
 
         return result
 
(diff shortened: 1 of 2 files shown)


aiocache/decorators.py
```diff
@@ -72,14 +72,19 @@ def __call__(self, f):
         return wrapper
 
     async def decorator(self, f, *args, **kwargs):
+        disable_read = kwargs.pop('aiocache_disable_read', False)
+        disable_write = kwargs.pop('aiocache_disable_write', False)
```
argaen: Put explicitly in the `function` signature please

argaen: Also, add docs about those two new parameters in the class docstring or in `decorators.rst`

tests/ut/test_decorators.py
```diff
@@ -94,6 +94,37 @@ def test_get_cache_key_with_key_builder(self, decorator):
         assert decorator.cache.set.call_count == 0
         assert stub.call_count == 0
 
+    @pytest.mark.asyncio
+    async def test_disable_read(self, decorator, decorator_call):
+        decorator.cache.get = CoroutineMock(return_value=1)
```
argaen: Why do you need this if its not being called?

tests/ut/test_decorators.py
```diff
@@ -434,6 +465,38 @@ def test_get_cache_keys_with_key_builder(self, decorator):
         with pytest.raises(Exception):
             assert await decorator_call(keys=[])
 
+    @pytest.mark.asyncio
+    async def test_disable_read(self, decorator, decorator_call):
+        decorator.cache.multi_get = CoroutineMock(return_value=[1, 2])
```
argaen: same, not needed because its not being called?

argaen:
> Not convinced about the naming (too long?), I'm open to suggestions.
Yeah I thought the same... Maybe `cache_read` and `cache_write` and set them to `True` by default?

tests/ut/test_decorators.py
jcugat: Probably from copy&pasting, removed

tests/ut/test_decorators.py
jcugat: Removed

aiocache/decorators.py
jcugat: Done

Issue #387, opened 2018-03-12
Allow all parameters in the cache config, and override them in specific usages


jcugat:
When setting the cache config, I'd like to set there default values (for TTL and namespace for example):

```py
caches.set_config({
    'default': {
        'cache': "aiocache.SimpleMemoryCache",
        'namespace': 'my_namespace',
        'ttl': 60,
    },
})
```
From that point, all the usages of the `default` cache would use the namespace and ttl set, unless passed explicitly in decorators and calls:

```py
@cached(alias='default')
async def something_cached(self):
    # This would cache with the default parameters defined in the config
    return await do_stuff()

@cached(alias='default', namespace='another_namespace', ttl=3600)
async def something_else_cached(self):
    # This would cache with the parameters defined in the decorator instead of the config
    return await do_stuff()

# Same for these
caches.get('default').get('foo', default=False, namespace='bar_namespace')
caches.get('default').set('foo', True, namespace='bar_namespace', ttl=86400)
```

argaen:
There are two things here. 

On one hand is the part of specifying all the parameters through config. This is already possible except for TTL but because there is no instance attribute for attribute, its just per function. If you want this, we could add ttl in the class constructor too.

On the other hand is the part of overriding with the specifics you pass in the decorator. I guess we could do that in https://github.com/argaen/aiocache/blob/master/aiocache/decorators.py#L59, if there is alias load the config and then create new cache with the new args if passed.

PR #405, opened 2018-05-26
Added ttl parameter in cache class

jcugat:
As discussed in #387 `ttl` was the only parameter not allowed by default in the cache class. Caches instances that have this parameter in the constructor will use it by default in all the calls, unless a specific parameter is passed.

### DIFF
--- a/aiocache/base.py
+++ b/aiocache/base.py
@@ -9,7 +9,7 @@ from aiocache import serializers
 
 logger = logging.getLogger(__file__)
 
-sentinel = object()
+SENTINEL = object()
 
 
 class API:
@@ -93,17 +93,17 @@ class BaseCache:
         list.
     :param namespace: string to use as default prefix for the key used in all operations of
         the backend. Default is None
-    :param ttl: int the expiration time in seconds to use as a default in all operations of
-        the backend. It can be overriden in the specific calls.
     :param key_builder: alternative callable to build the key. Receives the key and the namespace as
         params and should return something that can be used as key by the underlying backend.
     :param timeout: int or float in seconds specifying maximum timeout for the operations to last.
         By default its 5. Use 0 or None if you want to disable it.
+    :param ttl: int the expiration time in seconds to use as a default in all operations of
+        the backend. It can be overriden in the specific calls.
     """
 
     def __init__(
             self, serializer=None, plugins=None,
-            namespace=None, ttl=None, key_builder=None, timeout=5):
+            namespace=None, key_builder=None, timeout=5, ttl=None):
         self.timeout = timeout
         self.namespace = namespace
         self.ttl = ttl
@@ -135,7 +135,7 @@ class BaseCache:
     @API.aiocache_enabled(fake_return=True)
     @API.timeout
     @API.plugins
-    async def add(self, key, value, ttl=sentinel, dumps_fn=None, namespace=None, _conn=None):
+    async def add(self, key, value, ttl=SENTINEL, dumps_fn=None, namespace=None, _conn=None):
         """
         Stores the value in the given key with ttl if specified. Raises an error if the
         key already exists.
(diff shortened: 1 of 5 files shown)


aiocache/decorators.py
```diff
@@ -105,7 +106,10 @@ def _key_from_args(self, func, args, kwargs):
 
     async def set_in_cache(self, key, value):
         try:
-            await self.cache.set(key, value, ttl=self.ttl)
+            kwargs = {}
+            if self.ttl is not sentinel:
+                kwargs['ttl'] = self.ttl
```
argaen: why do we need to replicate the logic here? This is already done in the class so if we just propagate `self.ttl` here (having `sentinel` as default value) it should be enough right?

aiocache/decorators.py
```diff
@@ -291,8 +295,11 @@ def get_cache_keys(self, f, args, kwargs):
 
     async def set_in_cache(self, result, fn_args, fn_kwargs):
         try:
+            kwargs = {}
+            if self.ttl is not sentinel:
+                kwargs['ttl'] = self.ttl
```
argaen: same as before

tests/ut/test_base.py
```diff
@@ -231,6 +231,93 @@ def test_alt_build_key(self):
...
+        base_cache.ttl = None
+
+    @pytest.mark.asyncio
+    async def test_add_ttl_default(self, set_test_ttl, base_cache):
+        base_cache._add = CoroutineMock()
+
+        await base_cache.add(pytest.KEY, "value")
+
+        assert base_cache._add.call_count == 1
+        assert base_cache._add.call_args[1]['ttl'] == 10
```
argaen: Lets use `assert_called_once_with`, more compact

tests/ut/test_base.py
```diff
@@ -231,6 +231,93 @@ def test_alt_build_key(self):
         cache = BaseCache(key_builder=lambda key, namespace: 'x')
         assert cache.build_key(pytest.KEY, 'namespace') == 'x'
 
+    @pytest.fixture
+    def set_test_ttl(self, base_cache):
+        base_cache.ttl = 10
+        yield
+        base_cache.ttl = None
```
argaen: You don't need to reset the ttl to `None` every time. The `base_cache` fixture is per test so this fixture is not needed, just set the `ttl` of `base_cache` directly in the test whenever you need it

tests/ut/test_base.py
```diff
@@ -231,6 +231,93 @@ def test_alt_build_key(self):
...
+        assert base_cache._add.call_args[1]['ttl'] == 20
+
+    @pytest.mark.asyncio
+    async def test_add_ttl_none(self, set_test_ttl, base_cache):
+        base_cache._add = CoroutineMock()
+
+        await base_cache.add(pytest.KEY, "value", ttl=None)
+
+        assert base_cache._add.call_count == 1
+        assert base_cache._add.call_args[1]['ttl'] is None
```
argaen: I think we are missing the case where we don't pass `ttl` explicitly and its not set either at he class too

aiocache/base.py
```diff
@@ -9,6 +9,8 @@
 
 logger = logging.getLogger(__file__)
 
+sentinel = object()
```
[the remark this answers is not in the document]
jcugat: Done

aiocache/base.py
```diff
@@ -99,9 +103,10 @@ class BaseCache:
 
     def __init__(
             self, serializer=None, plugins=None,
-            namespace=None, key_builder=None, timeout=5):
+            namespace=None, ttl=None, key_builder=None, timeout=5):
```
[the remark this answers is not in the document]
jcugat: Done

aiocache/decorators.py
jcugat: I wanted to avoid having to update all the tests to use `SENTINEL` when mocking the call. Updated.

aiocache/decorators.py
jcugat: Updated

tests/ut/test_base.py
jcugat: Done

tests/ut/test_base.py
jcugat: Done

tests/ut/test_base.py
jcugat: Added

PR #415, opened 2018-07-25
Update cached_decorator.py

liulwx:
assert  bool don't need True


argaen:
Hi @liulwx thanks for your contribution. I know it seems redundant but I want to show in the example explicitly that the `exists` operation returns `True`. Sometimes it's dangerous not asserting explicitly for a specific value so I prefer leaving it as it is :)

PR #424, opened 2018-11-10
Cancel the previous ttl timer if exists when setting a new value in the in-memory cache

minhtule:
This PR is branched out from the v0.10.0 version bump [commit](https://github.com/argaen/aiocache/commit/f4c64c09a46bb5a8987e78462b09d15e19b17fbe) because there is some change on `master` (e.g. https://github.com/argaen/aiocache/commit/634348f40ce8caa01c7c35010acf32d8c3e17ba6) that is not backward compatible with `0.10.0`.


tests/acceptance/test_base.py
```diff
@@ -150,6 +150,17 @@ class TestMemoryCache:
...
+    async def test_set_cancel_previous_ttl_handle(self, memory_cache):
+        await memory_cache.set(pytest.KEY, "value", ttl=0.1)
+
+        await asyncio.sleep(0.05)
+        assert await memory_cache.get(pytest.KEY) == "value"
+        await memory_cache.set(pytest.KEY, "new_value", ttl=0.1)
+
+        await asyncio.sleep(0.05)
+        assert await memory_cache.get(pytest.KEY) == "new_value"
+
```
argaen: Hey @minhtule lets move this test to `TestCache` and use the `cache` fixture. This way we ensure all backends behave in a consistent way :)

minhtule: That makes sense. I've moved the test there.

minhtule:
@argaen I'm not sure why the 2 checks are still pending. Could you please take a look? Also I notice that the `TestCachedStampede.test_cached_stampede` test can randomly fail like in [this build](https://travis-ci.org/argaen/aiocache/jobs/453978932) (I saw it before too).

argaen:
> @argaen I'm not sure why the 2 checks are still pending. Could you please take a look?
Yeah.. it happened to me sometimes too, if the build passes correctly (I've resync with master) I'll merge the changes

> Also I notice that the TestCachedStampede.test_cached_stampede test can randomly fail like in this build (I saw it before too).
I know... It's been there for a while but haven't dedicated time to investigate TBH. Happy to accept PRs :D

argaen:
Merged, tomorrow I'll do a bugfix release, thanks :)

minhtule:
Thank you @argaen !

argaen:
@minhtule released yesterday with 0.10.1

minhtule:
Got it. Thank you @argaen !

PR #430, opened 2018-12-31
Add Cache factory

argaen:
Adds `aiocache.Cache` as a factory for creating different caches. Previously you would have to import the specific backend as
```
from aiocache import RedisCache

cache = RedisCache
```

now

```
from aiocache import Cache

cache = Cache(Cache.REDIS)
```

it just adds a level of indirection but will help on adding/centralizing features like creating a cache from a resource url, creating from a string for the config, etc. Also recommending this approach to create caches from now on


tomchristie:
Yup good stuff! I guess ultimately I'd prefer `Cache("redis://localhost/0")` over `Cache.from_url("redis://localhost/0")` but either way this change helps make that possible.

👍

PR #432, opened 2019-01-04
Feature #428/cache from url

### LINKED ISSUE #428: API to instantiate a cache from a resource url
From https://github.com/encode/starlette/issues/252.

Let's add an API to create caches from resource urls. First thing that comes to my mind is add support in `caches.create` call to do that.

argaen:
This PR adds the `Cache.from_url` method to instantiate caches from a specific resource url.

Couple of things:
- password is passed through queryparams. According to IANA proposal password is passed as the user info param. However user doesn't make sense because Redis authentication only goes with password.
- maybe it would make sense for redis to have the `db` as a path param instead of having it as a query param. However its the only one so I think its better to have it as queryparam like all the other things you can configure that then get propagated to the underlying client

Docs:
screenshot from 2019-01-05 21-53-47

Closes #428

### DIFF
--- a/aiocache/backends/memcached.py
+++ b/aiocache/backends/memcached.py
@@ -118,6 +118,10 @@ class MemcachedBackend:
     async def _close(self, *args, _conn=None, **kwargs):
         await self.client.close()
 
+    @classmethod
+    def parse_uri_path(self, path):
+        return {}
+
 
 class MemcachedCache(MemcachedBackend, BaseCache):
     """
--- a/aiocache/backends/memcached.py
+++ b/aiocache/backends/memcached.py
@@ -10,7 +10,7 @@ class MemcachedBackend:
         super().__init__(**kwargs)
         self.endpoint = endpoint
         self.port = port
-        self.pool_size = pool_size
+        self.pool_size = int(pool_size)
         self._loop = loop
         self.client = aiomcache.Client(
             self.endpoint, self.port, loop=self._loop, pool_size=self.pool_size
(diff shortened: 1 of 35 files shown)


aiocache/backends/memcached.py
```diff
@@ -10,7 +10,7 @@ def __init__(self, endpoint="127.0.0.1", port=11211, pool_size=2, loop=None, **k
         super().__init__(**kwargs)
         self.endpoint = endpoint
         self.port = port
-        self.pool_size = pool_size
+        self.pool_size = int(pool_size)
```
argaen: unrelated with PR but querystring from `from_url` method always parses args as strings. This is helpful anyway as a standalone feature

tomchristie:
> maybe it would make sense for redis to have the db as a path param instead of having it as a query param. However its the only one so I think its better to have it as queryparam like all the other things you can configure that then get propagated to the underlying client

Indeed. Other versions I've seen of redis URLs treat the path as the database component, which is consistent with eg. how database URLs are used. 

So: `redis://[:password@]host[:port][/db-number][?option=value]`

Eg: see https://www.npmjs.com/package/redis-url

tomchristie:
Also *provisionally* registered with IANA: https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml

argaen:
> Indeed. Other versions I've seen of redis URLs treat the path as the database component, which is consistent with eg. how database URLs are used.

Yeah and aioredis also accepts that. I was just being lazy because this means implementation of this feature will be backend specific then :(. Will update 😄

argaen:
Updated, @tomchristie let me know what you think :)

tomchristie:
Looks great!

PR #434, opened 2019-01-08
Improve typehints in lock.py

aviv-ebates:
- The constructors take an instance of type `BaseCache`, not `Type[]`
- Mypy doesn't consider `True` to be a type.

### DIFF
--- a/aiocache/lock.py
+++ b/aiocache/lock.py
@@ -1,7 +1,7 @@
 import asyncio
 import uuid
 
-from typing import Type, Union, Any
+from typing import Union, Any
 
 from aiocache.base import BaseCache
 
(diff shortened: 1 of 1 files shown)


argaen:
thanks @aviv-ebates! can you fix the syntax errors? no need to import the `Type` class anymore

aviv-ebates:
hey @argaen , I've fixed the `import`, but it kinda looks like the tests are broken with `unknown hook 'pytest_namespace'`....

argaen:
Oh wow, seems tests started failed since yesterday, probably a rogue dependency for testing. Will check and then rebase this branch with master then to see if build passes

aviv-ebates:
thanks!

argaen:
thanks for the contribution! :)

PR #443, opened 2019-02-06
Catch KeyError when the lock is already released

minhtule:
When there are many calls waiting, there could be a race condition when releasing the lock and I encounter `KeyError` when trying to pop the `asyncio.Event`.

### DIFF
--- a/aiocache/lock.py
+++ b/aiocache/lock.py
@@ -94,7 +94,10 @@ class RedLock:
     async def _release(self):
         removed = await self.client._redlock_release(self.key, self._value)
         if removed:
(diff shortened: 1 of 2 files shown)


argaen:
Hey @minhtule, do you have a snippet to reproduce this? with what backend was that? Supposedly the `_redlock_release` methods should be free of race conditions

minhtule:
@argaen I'm sorry for the late response. I'll try to find some time to make a reproducible environment for this issue.

minhtule:
Closing this since I won't be able to work on the demo code 😞

argaen:
Pity, if you run into it please feel free to reopen with something to reproduce

PR #446, opened 2019-03-07
Extend and improve usage of Cache class

argaen:
Trying to push usage of `Cache.MEMORY` accessor like instead of accessing the constructor directly

### DIFF
--- a/aiocache/backends/memcached.py
+++ b/aiocache/backends/memcached.py
@@ -118,10 +118,6 @@ class MemcachedBackend:
     async def _close(self, *args, _conn=None, **kwargs):
         await self.client.close()
 
-    @classmethod
-    def parse_uri_path(self, path):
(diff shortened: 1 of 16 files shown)


aiocache/backends/memcached.py
```diff
@@ -142,10 +138,16 @@ class MemcachedCache(MemcachedBackend, BaseCache):
     :param pool_size: int size for memcached connections pool. Default is 2.
     """
 
+    NAME = "memcached"
+
     def __init__(self, serializer=None, **kwargs):
         super().__init__(**kwargs)
         self.serializer = serializer or JsonSerializer()
 
+    @classmethod
```
argaen: This is from before, not sure why I did put the method in the backend class but backend should contain methods only related to the client implementation and this is not

PR #458, opened 2019-06-18
Add python38 in the pipeline

argaen:


argaen:
Two tests failing 
https://travis-ci.org/argaen/aiocache/jobs/565782461

PR #474, opened 2019-12-27
Add if statement to _clear method in redis.py

AbuSM:


argaen:
Thanks @AbuSM. Could you please add a unit test below https://github.com/argaen/aiocache/blob/master/tests/ut/backends/test_redis.py#L281 checking the delete function is not called if no keys are returned?

AbuSM:
> Thanks @AbuSM. Could you please add a unit test below https://github.com/argaen/aiocache/blob/master/tests/ut/backends/test_redis.py#L281 checking the delete function is not called if no keys are returned?

Hi @argaen, sorry for delayed reply, as I understood you correctly and as I know ```key``` must be passed to ```delete``` method else it will give TypeError for missing argument. So I think there is no need to write unit test on this, please correct me if I missed smth!

argaen:
Sorry @AbuSM I mean below the `test_clear` (https://github.com/argaen/aiocache/blob/master/tests/ut/backends/test_redis.py#L286). Test should be something like:

```
 @pytest.mark.asyncio
    async def test_clear_when_no_keys(self, redis, redis_connection):
        redis_connection.keys.return_value = []
        await redis._clear("nm")
        redis_connection.delete.assert_not_called()
```

AbuSM:
> Sorry @AbuSM I mean below the `test_clear` (https://github.com/argaen/aiocache/blob/master/tests/ut/backends/test_redis.py#L286). Test should be something like:
> 
> ```
>  @pytest.mark.asyncio
>     async def test_clear_when_no_keys(self, redis, redis_connection):
>         redis_connection.keys.return_value = []
>         await redis._clear("nm")
>         redis_connection.delete.assert_not_called()
> ```

AbuSM:
@argaen I'm new in writing tests((, however I've found test for empty namespace a few lines below of your linked code:
```
@pytest.mark.asyncio
    async def test_clear_no_namespace(self, redis, redis_connection):
        await redis._clear()
        assert redis_connection.flushdb.call_count == 1
```
https://github.com/argaen/aiocache/blob/24174d02b55773026f80741ceadf1c24b6258f17/tests/ut/backends/test_redis.py#L291

So, I think there is no need for writing unit test eventually)
Sorry for my omissions if they exist...

argaen:
Closing in favor of https://github.com/argaen/aiocache/pull/478 where I added test

Issue #475, opened 2019-12-30
Use decorator on aiohttp route and not on a specific function


DaRoee:
Currently, the usage for caching routes is to decorate a function the route is using, though what can be really cool is caching the whole route response.
Using it as a decorator for the route is making the response fail...

I can implement it if you want, should be fairly quick and it's pretty much a requirement for my project :)

argaen:
What's the failure? Could you post an example traceback? I'm also interested on what would be the changes required for this so feel free to open a quick PR and maybe we can discuss

DaRoee:
My traceback is pretty generic, coming from nginx about a closed connection while reading the response header. 
The real problem can be understood here: https://docs.aiohttp.org/en/stable/faq.html#how-do-i-make-sure-my-custom-middleware-response-will-behave-correctly
We can't really reuse our cached response, so I think the solution will be to recreate aiohttp responses everytime the cache is called.
I'll do a PR later today/tomorrow, where we can further discuss about the solution :)

argaen:
Closing as per https://github.com/argaen/aiocache/pull/476 discussion

PR #476, opened 2019-12-31
Recreate aiohttp response since it can't be reused for multiple requests (#475)

DaRoee:
It's a simple solution, and I'm unaware if sanic and tornado have the same limitations...
But you can now decorate and cache a whole route:
```
@cached(ttl=60, key="random_key")
async def get(self, *args, **kwargs):
...
```

EDIT:
Travis checks made me realise this is not a good solution since it creates a dependency for aiohttp that didn't really exist so far...
I think the best solution here is to inherit from cached and override`get_from_cache`with that way you want it. I can update the examples if you want me, but I'm not sure there are any changes necessary to the code...

### DIFF
--- a/aiocache/decorators.py
+++ b/aiocache/decorators.py
@@ -3,6 +3,8 @@ import inspect
 import functools
 import logging
 
+from aiohttp.web import Response as AIOResponse
+
 from aiocache import Cache, caches
 from aiocache.base import SENTINEL
 from aiocache.lock import RedLock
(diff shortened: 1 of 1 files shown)


argaen:
I'm happy for you to add an example on how to cache aiohttp handlers with `cached`. Agree this is not a pretty solution and each user should inherit and customise as needed depending on the framework that is being used :)

Issue #482, opened 2020-04-10
JsonSerializer & SimpleMemoryBackend: get after increment raises TypeError


Sergei-Rudenkov:
- int type is added into dict in [SimpleMemoryBackend.increment](https://github.com/argaen/aiocache/blob/0.11.1/aiocache/backends/memory.py#L60) 
- [JsonSerializer.loads](https://github.com/argaen/aiocache/blob/0.11.1/aiocache/serializers/serializers.py#L151) tries to load int value.

As result: TypeError

Sergei-Rudenkov:
pull request with fix: https://github.com/argaen/aiocache/pull/483

Dreamsorcerer:
Problem is that increment seems to be an operation that should be done on the backend efficiently. This requires storing the value as an int in the backend.

If you are using a serialiser that changes the values (e.g. JSON converts to str stored in the backend), then this obviously won't work.

We should put a warning or something in the docs, and maybe see if we can disable the `increment()` method if an inappropriate serialiser is being used (or maybe a fallback method could be used which just does `set(key, get(key) + 1)`.

PR #483, opened 2020-04-10
#482. fix issue: get after increment raises TypeError in JsonSerializer

Sergei-Rudenkov:


argaen:
This makes tests fail. Please include a snippet to reproduce the error and update tests accordingly

PR #487, opened 2020-04-16
code objects analysis replaced with inspect

Trizalio:
`multi_cached` works incorrectly with decorated functions, called with `keys_from_attr` value passed as positional argument, example:

```python
import asyncio
import functools
from typing import List, Dict

    return {id_: id_ ** 2 for id_ in ids}


def simple_decorator(f):
    @functools.wraps(f)
    async def wrapper(*args, **kwargs):
        return await f(*args, **kwargs)
    return wrapper


cached_foo = multi_cached("ids", namespace="main")(foo)
cached_wrapped_foo = multi_cached("ids", namespace="main")(simple_decorator(foo))


async def amain():
    await cached_foo(ids=[1])  # ok
    await cached_foo([1])  # ok
    await cached_wrapped_foo(ids=[1])  # ok
    await cached_wrapped_foo([1])  # EXCEPTION

if __name__ == "__main__":
    asyncio.run(amain())
```
This problem's reason is analysis of code object of function passed to `multi_cached` decorator.   
To properly get arguments from function decorated with `functools.wraps`, it's required to get signature of inner function. Which can be accessed by `__wrapped__` attribute, or by `inspect.signature` function, which goes recursively through `__wrapped__` attributes till core function.   
I suggest to used `inspect.signature` instead of direct code obejects inteaction.

### DIFF
--- a/aiocache/decorators.py
+++ b/aiocache/decorators.py
@@ -207,13 +207,11 @@ def _get_cache(cache=Cache.MEMORY, serializer=None, plugins=None, **cache_kwargs
 
 
 def _get_args_dict(func, args, kwargs):
-    defaults = {
-        arg_name: arg.default
-        for arg_name, arg in inspect.signature(func).parameters.items()
-        if arg.default is not inspect._empty  # TODO: bug prone..
-    }
-    args_names = func.__code__.co_varnames[: func.__code__.co_argcount]
-    return {**defaults, **dict(zip(args_names, args)), **kwargs}
+    signature = inspect.signature(func)
+    incoming_args = signature.bind(*args, **kwargs)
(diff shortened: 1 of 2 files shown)


Trizalio:
i`m quite new in coverage checks, can someone tell me, how is `serializers.py` file affected by my commit?

argaen:
Some tests are failing so it can be because of this as a side effect. Don't check coverage until the build passes

Dreamsorcerer:
Hey, sorry for the delay, if you can merge master and get this running again, then we can look at merging this.

If I understand correctly, the test is only affected by function signatures, so you should also move the test to tests/ut/test_decorators.py and use a mock_cache instead.

PR #489, opened 2020-04-17
reversed "update" method in multi_cache to overwrite old values with new

Trizalio:
If wrapped function can return dict with more keys, than required there is a case, in which new value will be replaced by old, example:
```python
import asyncio
from typing import List, Dict

@multi_cached("ids", namespace="main")
    biggest_value = max(ids)
    result = {id_: biggest_value for id_ in range(biggest_value + 1)}
    print(f'foo result: {result}')
    return result


async def amain():
    print(await foo(ids=[1]))
    print(await foo(ids=[0, 2]))
    print(await foo(ids=[0]))

if __name__ == "__main__":
    asyncio.run(amain())
```
will be printed:
```
foo result: {0: 1, 1: 1}
{0: 1, 1: 1}
foo result: {0: 2, 1: 2, 2: 2}
{0: 1, 1: 2, 2: 2}
{0: 1}

```
As we can see, on second call of `foo` value for `1` changed, but it was overwritten by cache value.

In my case i must return aggregated values based on values from two distinct bigdata storages. Based on requested keys i can predict future requests, so i'd like to preload values for predicted keys with keys passed to my funstion and save them all in aiocache

### DIFF
--- a/aiocache/decorators.py
+++ b/aiocache/decorators.py
@@ -325,8 +325,9 @@ class multi_cached:
         else:
             kwargs[self.keys_from_attr] = missing_keys
 
-        result = await f(*new_args, **kwargs)
-        result.update(partial)
+        result = partial
+        new_items = await f(*new_args, **kwargs)
+        result.update(new_items)
 
         if cache_write:
             if aiocache_wait_for_write:


argaen (changes requested):
Thanks for this @Trizalio!

We would need to add tests in order to merge this

Dreamsorcerer:
Looks reasonable, but as mentioned, please add a test (after merging master). Test probably should go in tests/ut/test_decorators.py

PR #492, opened 2020-05-19
Fix TypeError on import all from aiocache

nkoshell:
Python 3.7.7

```python
from aiocache import *
```

Result:
```
TypeError: Item in aiocache.__all__ must be str, not type
```


nkoshell:
Tests failed, but not due to changes made by me

argaen:
Thanks!

PR #497, opened 2020-06-13
ENH: create default_ttl variable

rspadim:
This commit allow a default behavior to cache set functions

We could set this at config json, for example:
```
"cache": {
    "cache": "aiocache.SimpleMemoryCache",
    "default_ttl": 300,
    "serializer": {
        "class": "aiocache.serializers.PickleSerializer"
    }
},
```

### DIFF
--- a/aiocache/backends/memcached.py
+++ b/aiocache/backends/memcached.py
@@ -8,6 +8,8 @@ from aiocache.serializers import JsonSerializer
 class MemcachedBackend:
     def __init__(self, endpoint="127.0.0.1", port=11211, pool_size=2, loop=None, **kwargs):
         super().__init__(**kwargs)
+        # TODO: this is a super() option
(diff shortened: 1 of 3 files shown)


rspadim:
this is very helpfull to devops

with a single configuration file we can set a standard ttl to many caches

this remove the developer from hardcoding ttl values or use of environment variables in each new project

it's a good practice, a single place to set caches

developer can use a `caches.get("cache system name")` instead of harcoded value `cache.set(ttl=999)` or `cache.set(ttl=os.environ['some var'])`

rspadim:
there's a ttl param that already do this =) sorry about wrong mr

https://aiocache.readthedocs.io/en/latest/caches.html#aiocache.base.BaseCache

ttl – int the expiration time in seconds to use as a default in all operations of the backend. It can be overriden in the specific calls.
--

PR #503, opened 2020-07-22
fix: deletion of items that don't implement truthiness

JeppeKlitgaard:
Example: pandas.DataFrame or pandas.Series.

The current memory backend implementation evaluates truthiness on deletion from the cache.

This works out most of the time but results in an error when caching pandas.Dataframe or pandas.Series objects.

### DIFF
--- a/aiocache/decorators.py
+++ b/aiocache/decorators.py
@@ -183,12 +183,20 @@ class cached_stampede(cached):
         super().__init__(**kwargs)
         self.lease = lease
 
-    async def decorator(self, f, *args, **kwargs):
+    async def decorator(self,
+        f,
+        *args,
+        cache_read=True,
+        cache_write=True,
+        aiocache_wait_for_write=True,
+        **kwargs
+    ):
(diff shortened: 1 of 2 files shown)


argaen:
Hey @JeppeKlitgaard would you mind adding tests for this please?

aiocache/decorators.py
```diff
@@ -183,12 +183,20 @@ def __init__(self, lease=2, **kwargs):
         super().__init__(**kwargs)
         self.lease = lease
 
-    async def decorator(self, f, *args, **kwargs):
+    async def decorator(self,
+        f,
+        *args,
+        cache_read=True,
+        cache_write=True,
+        aiocache_wait_for_write=True,
```
argaen, line 191: these changes seem unrelated with what you propose, can you remove them please?

bbelyeu: Agreed. The cache_read and cache_write should be a new issue and a different MR if you'd like to propose that functionality. It's a different discussion (feature add) as opposed to the original bug fix.

PR #507, opened 2020-09-24
Fix. Add kwargs to plugin methods.

dmtr:
It fixes next problem: if for example namespace parameter is passed to `get` method,  then `post_get` method raises unexpected keyword argument error.

### DIFF
--- a/tests/acceptance/test_plugins.py
+++ b/tests/acceptance/test_plugins.py
@@ -1,7 +1,6 @@
 import pytest
-
-from aiocache.plugins import HitMissRatioPlugin, TimingPlugin
 from aiocache.backends.memory import SimpleMemoryBackend
+from aiocache.plugins import HitMissRatioPlugin, TimingPlugin
 
 
 class TestHitMissRatioPlugin:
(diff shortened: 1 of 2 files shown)


argaen:
Hey @dmtr can you please add a snippet to reproduce this and also tests for the changes?

dmtr:
Hi!
This code reproduces the bug.
```
import aiocache
import asyncio

config = {'default': {'cache': 'aiocache.RedisCache', 'endpoint': 'localhost', 'port': 6379, 'timeout': 5, 'serializer': {'class': 'aiocache.serializers.JsonSerializer'}, 'plugins': [{'class': 'aiocache.plugins.HitMissRatioPlugin'}]}}

aiocache.caches.set_config(config)
cache = aiocache.caches.get("default")
loop = asyncio.get_event_loop()

loop.run_until_complete(cache.set("B", 2, namespace="test"))
loop.run_until_complete(cache.get("B", namespace="test"))
```

It raises error
```
~/.pyenv/versions/3.7.4/envs/exness/lib/python3.7/site-packages/aiocache/base.py in _plugins(self, *args, **kwargs)
     78             for plugin in self.plugins:
     79                 await getattr(plugin, "post_{}".format(func.__name__))(
---> 80                     self, *args, took=end - start, ret=ret, **kwargs
     81                 )
     82             return ret

TypeError: post_get() got an unexpected keyword argument 'namespace'
```

bbelyeu:
Looks good to me. Thanks for the contribution @dmtr !

tests/acceptance/test_plugins.py
```diff
@@ -55,6 +54,21 @@ async def test_multi_get_hit_miss_ratio(self, memory_cache, data, ratio):
             == len(hits) / memory_cache.hit_miss_ratio["total"]
         )
 
+    @pytest.mark.asyncio
+    @pytest.mark.parametrize(
+        "key,value,namespace", [("a", 1, None), ("b", 2, "test")],
+    )
+    async def test_set_and_get(self, memory_cache, key, value, namespace):
```
argaen, line 61: the case of not having a namespace is already tested above so we are good for the first case. Let's leave only one test where we pass the namespace explicitly and then we are ready to merge :)

argaen:
@dmtr once you've changed the test as I mentioned please remember to do `make format` before pushing. Otherwise linting will fail (that's why the build is failing)

PR #508, opened 2020-09-28
fix: deletion of items that don't implement truthiness

JeppeKlitgaard:
Example: pandas.DataFrame or pandas.Series.

Now features a test as well.

I believe the build will fail, but this is unrelated to this PR.


argaen (approved):
Thanks for this!

PR #509, opened 2020-09-28
fix: lint with local venv

JeppeKlitgaard:
Currently the `make lint` command will lint all installed packages
if using a local venv.

This commit changes the flake8 linter to be specific
(like the black formatter already is)


bbelyeu:
Rebased. Change looks good to me.

argaen (approved):

PR #510, opened 2020-09-28
fix: cache control arguments for cached_stampede

JeppeKlitgaard:
`cached_stampede` now features the same cache control arguments
as the `cached` decorator


argaen:
@JeppeKlitgaard thanks for this, mind adding tests? There should be ones for the `cached` decorator that you can just copy paste

Dreamsorcerer:
As mentioned, tests please, otherwise we'll close this soon.

PR #513, opened 2020-10-14
Add python3.8 support and remove python3.5

argaen:
Aioredis python3.8 support starts at version 1.3.0 at least for the functionalities the package uses so restricting that

### DIFF
--- a/aiocache/decorators.py
+++ b/aiocache/decorators.py
@@ -338,7 +338,7 @@ class multi_cached:
 
     def get_cache_keys(self, f, args, kwargs):
         args_dict = _get_args_dict(f, args, kwargs)
-        keys = args_dict[self.keys_from_attr] or []
+        keys = args_dict.get(self.keys_from_attr, []) or []
         keys = [self.key_builder(key, f, *args, **kwargs) for key in keys]
 
         args_names = f.__code__.co_varnames[: f.__code__.co_argcount]
--- a/aiocache/serializers/serializers.py
+++ b/aiocache/serializers/serializers.py
@@ -99,6 +99,10 @@ class PickleSerializer(BaseSerializer):
 
     DEFAULT_ENCODING = None
 
+    def __init__(self, *args, protocol=pickle.DEFAULT_PROTOCOL, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.protocol = protocol
+
     def dumps(self, value):
         """
         Serialize the received value using ``pickle.dumps``.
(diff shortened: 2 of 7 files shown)


aiocache/decorators.py
```diff
@@ -338,7 +338,7 @@ async def decorator(
 
     def get_cache_keys(self, f, args, kwargs):
         args_dict = _get_args_dict(f, args, kwargs)
-        keys = args_dict[self.keys_from_attr] or []
+        keys = args_dict.get(self.keys_from_attr, []) or []
```
argaen, line 341: inspect.signature changed and in new versions it parses the defaults of the function correctly

aiocache/serializers/serializers.py
```diff
@@ -99,14 +99,18 @@ class PickleSerializer(BaseSerializer):
 
     DEFAULT_ENCODING = None
 
+    def __init__(self, *args, protocol=pickle.DEFAULT_PROTOCOL, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.protocol = protocol
```
argaen, line 104: Adding ability for the PickleSerializer to have a protocol passed. By default in later versions it is 4 while in previous it is 3 which results in differences when dumping content

setup.py
```diff
@@ -28,16 +28,17 @@
...
+        'Programming Language :: Python :: 3.8',
         'Framework :: AsyncIO',
     ],
     packages=find_packages(),
     install_requires=None,
     extras_require={
         'redis:python_version<"3.7"': ['aioredis>=0.3.3'],
-        'redis:python_version>="3.7"': ['aioredis>=1.0.0'],
+        'redis:python_version>="3.8"': ['aioredis>=1.3.0'],
+        'redis:python_version>="3.7" and python_version<"3.8"': ['aioredis>=1.0.0'],
```
argaen, line 41: tests for python3.8 only pass starting aioredis>=1.3.0 so restricting this in the setup file

tests/ut/test_decorators.py
```diff
@@ -414,8 +414,7 @@ def test_get_cache_keys_empty_list(self, decorator):
         assert decorator.get_cache_keys(stub_dict, (), {"keys": []}) == ([], [], -1)
 
     def test_get_cache_keys_missing_kwarg(self, decorator):
-        with pytest.raises(KeyError):
-            assert decorator.get_cache_keys(stub_dict, (), {})
+        assert decorator.get_cache_keys(stub_dict, (), {}) == ([], [], -1)
```
argaen, line 417: this is a result of the change of inspect.signature now parsing default arguments correctly as it doesn't fail anymore

Issue #479, opened 2020-02-21
Calling 'clear' on namespaced SimpleMemoryCache instance clears entire underlying cache


edaniszewski:
When calling `clear` on an instance of a `SimpleMemoryCache` with a namespace assigned, it will clear the entire underlying cache instead of using the instance namespace. An example:

```py
import aiocache
import asyncio

# Define namespace constants
NS_ONE = 'ns.one'
NS_TWO = 'ns.two'

# Define cache references
cache_one = aiocache.SimpleMemoryCache(namespace=NS_ONE)
cache_two = aiocache.SimpleMemoryCache(namespace=NS_TWO)


async def do_something():
    # Add some data to CACHE_ONE, then get it and print it out
    # to verify it is there.
    await cache_one.set('foo', 'bar')
    val = await cache_one.get('foo')
    print(f'cache_one    k=foo v={val}')
    print(f'underlying cache: {cache_one._cache}')
    print('')

    # Call `clear` on cache_two. Since the cache reference is differentiated
    # by namespace, one would expect the clear to apply to the namespace of
    # the referenced cache.
    await cache_two.clear()

    # Get the value from cache_one again. It is not there because the
    # entire cache was cleared, not just the data under the cache_two namespace.
    val = await cache_one.get('foo')
    print(f'cache_one    k=foo v={val}')
    print(f'underlying cache: {cache_one._cache}')


if __name__ == '__main__':
    loop = asyncio.get_event_loop()
    loop.run_until_complete(do_something())
```

Where the output is:
```
cache_one    k=foo v=bar
underlying cache: {'ns.onefoo': 'bar'}

cache_one    k=foo v=None
underlying cache: {}
```

Specifying the namespace explicitly for the call to `clear` behaves as expected though
```diff
- await cache_two.clear()
+ await cache_two.clear(NS_TWO)
```
leading to the output
```
cache_one    k=foo v=bar
underlying cache: {'ns.onefoo': 'bar'}

cache_one    k=foo v=bar
underlying cache: {'ns.onefoo': 'bar'}
```

I believe this is related to this bit of code:
https://github.com/argaen/aiocache/blob/f811995e0632b46a4597cc07007c26f755d2ffa3/aiocache/base.py#L417-L431

Where the namespace is only ever passed in via parameter, never checking to see if the instance's `namespace` member is set.

PR #523, opened 2020-12-01
[WIP] Issue 479 - Calling 'clear' on namespaced SimpleMemoryCache instance clears entire underlying cache

jbondia:
## What do these changes do? Are there changes in behavior for the user?

These changes ensure the cache in `SimpleMemoryCache` and `RedisCache` is not wiped when `clear` is called without parameters but a namespace exists since it was specified at creation time.

`clear` in `SimpleMemoryCache` and `RedisCache` will wipe only the keys related to the specified namespace in the param or at instantiation time -> `clear` will not wipe the whole cache when called without params (unless it was not specified at creation time).

## Related issue number

https://github.com/aio-libs/aiocache/issues/479

## Checklist


aiocache/backends/memory.py
```diff
@@ -78,9 +78,10 @@ async def _delete(self, key, _conn=None):
         return self.__delete(key)
 
     async def _clear(self, namespace=None, _conn=None):
-        if namespace:
+        ns = namespace or self.namespace
```
argaen, line 81: let's keep the name of the original variable

```
namespace = namespace or self.namespace
```

aiocache/base.py
```diff
@@ -431,6 +431,14 @@ async def clear(self, namespace=None, _conn=None):
...
 
     async def _clear(self, namespace, _conn=None):
+        """
+        Clears the cache according to each type's implementation.
+        WARNING: if no namespace is specified when `clear` is called or when the cache instance
+        is created, it will clear all the namespaces in the cache and all nodes in Redis.
+
+        :param namespace: cache namespace to clear
+        raises: :`NotImplementedError` if the cache type does not implement this method
+        """
```
argaen, line 441: please update the docstring in `clear` method instead which is what appears in the docs: https://aiocache.readthedocs.io/en/latest/caches.html#aiocache.base.BaseCache.clear

tests/ut/backends/test_memory.py
```diff
@@ -187,8 +187,8 @@ async def test_clear_no_namespace(self, memory):
         SimpleMemoryBackend._handlers = "asdad"
         SimpleMemoryBackend._cache = "asdad"
         await memory._clear()
-        SimpleMemoryBackend._handlers = {}
-        SimpleMemoryBackend._cache = {}
+        assert {} == SimpleMemoryBackend._handlers
+        assert {} == SimpleMemoryBackend._cache
```
argaen, line 191: I like to follow the make sure X is equals to Y where Y is what you expect and X is the variable, keep it this way please :)

tests/ut/backends/test_memory.py
```diff
@@ -225,3 +225,21 @@ def test_default_serializer(self):
...
+    @pytest.mark.asyncio
+    async def test_clear_cache_self_namespace(self):
+        namespace_a = "a"
+        namespace_b = "b"
+        cache_a = SimpleMemoryCache(namespace=namespace_a)
+        cache_b = SimpleMemoryCache(namespace=namespace_b)
+
+        await cache_a.set("foo", "bar")
+        val = await cache_a.get("foo")
+        assert "bar" == val
```
argaen, line 238: why do we need this assert? there are tests for the `get` already

aiocache/backends/memory.py
jbondia: Will do this on the other file too

tests/ut/backends/test_memory.py
```diff
@@ -225,3 +225,21 @@ def test_default_serializer(self):
...
+
+        await cache_b.clear()
+        for key in cache_b._cache.keys():
+            assert not key.startswith(namespace_b)
```
argaen, line 242: let's not use loops if the cache is empty this test passes. I would just go with an assert the _cache value and that's it

tests/ut/backends/test_redis.py
```diff
@@ -377,3 +377,15 @@ def test_build_key_double_dot(self, set_test_namespace, redis_cache, namespace,
 
     def test_build_key_no_namespace(self, redis_cache):
         assert redis_cache.build_key(pytest.KEY, namespace=None) == pytest.KEY
+
+    @pytest.mark.asyncio
+    async def test_clear_cache_self_namespace(self, redis_pool):
```
argaen, line 382: why do we need two pools? Also this doesn't exactly check that the clear behaves as expected

tests/ut/backends/test_memory.py
argaen: ah wow... haha good one

tests/ut/backends/test_redis.py
jbondia: Ooooh true, I misunderstood the whole thing. Will re do

tests/ut/backends/test_memory.py
jbondia: Well, `cache_b._cache` can't be empty since it uses the same memory address for `_cache`, given how the backend is implemented `_cache` is a class variable.

I was trying to proof that all `b` keys were gone and the `a` ones were kept (before all of them would have been gone).

Do you have another suggestion on how to do this without a loop?

argaen: For unit tests I would follow same approach is already used for testing the clear method
image


then in acceptance I would do something like this:

```
await cache_a.set("a", 1)
await cache_b.set("b", 2)

await cache_b.clear()

assert cache_a.exists("a")
assert not cache_b.exists("b")
```

tests/ut/backends/test_redis.py
jbondia: Pushing this change now

Dreamsorcerer:
Think this has been resolved in #562 or another PR.

PR #529, opened 2021-01-25
Security Fix for Arbitrary Code Execution - huntr.dev

huntr-helper:
https://huntr.dev/users/B3EF has fixed the Arbitrary Code Execution vulnerability 🔨. Think you could fix a vulnerability like this?

Get involved at https://huntr.dev/

Q | A
Version Affected | ALL
Bug Fix | YES
Original Pull Request | https://github.com/418sec/aiocache/pull/1
Vulnerability README | https://github.com/418sec/huntr/blob/master/bounties/pip/aiocache/1/README.md

### 📊 Metadata *
Arbitrary Code execution Fix

### ⚙️ Description *
Asyncio cache manager for redis, memcached and memory, This library aims for simplicity over specialization.
### 💻 Technical Description *
untrusted loading of data by the pickle.load function leading to Arbitrary code execution.
### 🐛 Proof of Concept (PoC) *

Screenshot from 2021-01-07 18-36-19

### 🔥 Proof of Fix (PoF) *

Screenshot from 2021-01-07 18-38-39

### 👍 User Acceptance Testing (UAT)

Library tested successfully after the fix is being applied

### DIFF
--- a/aiocache/serializers/serializers.py
+++ b/aiocache/serializers/serializers.py
@@ -1,5 +1,21 @@
 import logging
 import pickle
+import io
+import builtins
+
+class RestrictedUnpickler(pickle.Unpickler):
+
+    def find_class(self, module, name):
+        """Only allow safe classes from builtins"""
+        if module == "builtins" and name in safe_builtins:
+            return getattr(builtins, name)
+        """Forbid everything else"""
+        raise pickle.UnpicklingError("global '%s.%s' is forbidden" %
+                                     (module, name))
+
+def restricted_loads(s):
+    """Helper function analogous to pickle.loads()"""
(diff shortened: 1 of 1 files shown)


Dreamsorcerer:
The "fix" looks like it would break even the demo in the README. The library seems to be designed to allow arbitrary objects and limiting it here does not make much sense to me.

I don't see any actual vulnerability either. The example only runs arbitrary code because the developer has created a class with a `__reduce__()` method which runs arbitrary code. I can't think of any real situation where this would actually allow arbitrary code to be executed.
    The library's purpose is also used for caching objects in the code, so it's not like we are loading untrusted serialised data. We are always unpickling something that was previously pickled in the user's code.

PR #533, opened 2021-03-11
Add parameter with dynamic_ttl_coroutine

mdczaplicki:
## What do these changes do?
Allows to use a coroutine to get a ttl.

No, only new functionality

## Related issue number
None

## Checklist


mdczaplicki:
I can update docs, but what do you think about it?

Dreamsorcerer:
Sorry for the delay. I'm not sure what the use case is for this (especially as you don't even pass any parameters to the function)?

If we did want this, I'd also reuse the `ttl` parameter with a `callable()` check or similar, given that you can't use `ttl` at the same time as this.

Dreamsorcerer:
I'm going to close this as it's not clear what the benefit is. Can be reopened if you want to explain the use cases.

PR #538, opened 2021-04-22
Allow user to bypass the cache when calling the @cached-decorated coroutine

theomart:
## What do these changes do?

It allows the user of the cached decorator to choose a keyword argument he can pass to the decorated function to easily bypass the cache.

If bypass_kwarg is not specified, nothing changes for the user. The user can choose a different keyword for each decorated function to conveniently  avoid conflicts with actual arguments passed to the non-cache related logic.

```python
import aiocache
import asyncio

@aiocache.cached(bypass_kwarg='refresh')
async def decorated_function(a):
    await asyncio.sleep(3)
    return a
    
loop = asyncio.get_event_loop()
print(loop.run_until_complete(decorated_function('test')))
print(loop.run_until_complete(decorated_function('test', refresh=False)))
print(loop.run_until_complete(decorated_function('test', refresh=True)))
```

## Checklist

### DIFF
--- a/aiocache/decorators.py
+++ b/aiocache/decorators.py
@@ -51,6 +51,9 @@ class cached:
     :param noself: bool if you are decorating a class function, by default self is also used to
         generate the key. This will result in same function calls done by different class instances
         to use different cache keys. Use noself=True if you want to ignore it.
+    :param bypass_kwargs: str the keyword argument containing a bool you want to pass to the decorated 
+        function to bypass the cache. For example, if ``bypass_kwargs='refresh'``, you can call
+        ``decorated_function(..., refresh=True, ...)`` to bypass the cache.
     """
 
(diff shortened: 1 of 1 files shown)


Dreamsorcerer:
Seems like a decent idea, let me know if you want to come back and finish this off.

I'm thinking it might be cleaner if we just add a method or something to the decorated object, so you'd just do: `my_func.refresh(*args)` or similar.

We could also just clear the cache beforehand, which looks like it can already be done with something like `my_func.cache.delete(key)`. With calculating the key being the difficult part, so a helper method to do that would still be good, so you just do `my_func.clear(*args)`.

Dreamsorcerer:
I've opened #625 for this idea. Let me know if you want to work on it.

Issue #543, opened 2021-07-30
redis doesn't work after aioredis package was updated


citramon:
Hi! 
We have a problem after new version [aioredis](https://github.com/aio-libs/aioredis-py/releases/tag/v2.0.0) was released.

```
AttributeError: module 'aioredis' has no attribute 'create_pool'
  File "my_project/apps/checks/services.py", line 66, in my_function
    if await self.cache.exists('key'):
  File "aiocache/base.py", line 61, in _enabled
    return await func(*args, **kwargs)
  File "aiocache/base.py", line 45, in _timeout
    return await asyncio.wait_for(func(self, *args, **kwargs), timeout)
  File "asyncio/tasks.py", line 442, in wait_for
    return fut.result()
  File "aiocache/base.py", line 75, in _plugins
    ret = await func(self, *args, **kwargs)
  File "aiocache/base.py", line 354, in exists
    ret = await self._exists(ns_key, _conn=_conn)
  File "aiocache/backends/redis.py", line 19, in wrapper
    pool = await self._get_pool()
  File "aiocache/backends/redis.py", line 219, in _get_pool
    self._pool = await aioredis.create_pool((self.endpoint, self.port), **kwargs)
```

AlekseiMarinichenko:
@citramon Use 

```
aiocache
aioredis>=1.3,<2.0
```
in your **requirements.txt** as a temporary fix.

aiocache is unmaintained and I do not think that someone will update it in a near feature

citramon:
> @citramon Use
> 
> ```
> aiocache
> aioredis>=1.3,<2.0
> ```
> 
> in your **requirements.txt** as a temporary fix.
> 
> aiocache is unmaintained and I do not think that someone will update it in a near feature

yep, i did it. Thanks. But I hope the maintainers will fix this someday :)

mattyweb:
@citramon It doesn't look like this project has maintainers. But it looks like it shouldn't be too difficult to fork to add support for aioredis 2.0. There's new syntax for the pool creation and updating that might do it.

It's a shame that this library isn't getting updated though. I'm not sure how aio-libs works but you'd think they'd be kept in sync with one another. Maybe one of the new aioredis maintainers would be gracious enough to take up ownership of this one too...

Krukov:
Hello, maybe It well be helpful for someone -  I got my own library like aiocache, but with supporting aioredis old and new version https://github.com/Krukov/cashews

samuelbailey123:
[...]

Issue #536, opened 2021-04-01
New release?


zware:
There are a few useful changes that are [not available in the latest 0.11.1 release](https://github.com/aio-libs/aiocache/compare/0.11.1...master), such as #478 (fixing Redis `clear` with namespace that happens to be empty), #470 (quieter imports), #514 (explicit support for Python 3.9).

I couldn't find documentation of release plans, so I figured I would just ask here :)

Thanks!

amatai:
Also interested to know about this plan. We are having to fork and use our own `master` branch for the redis clear fix.
Will be very helpful to have a new release.
Willing to help in this effort.

Mark90:
@Dreamsorcerer Should the Github CI build upon the tox/Makefile setup, or doesn't matter how it does the job as long as it lints tests and deploys?

Dreamsorcerer:
Doesn't need tox, just use a version matrix like in the other projects.

citramon:
@Dreamsorcerer, could you check [572](https://github.com/aio-libs/aiocache/pull/572)? It's for the last checkbox :)

nikolaevigor:
so is this project dead?

Dreamsorcerer:
Tests are running again, except for #585. If anyone has time to debug that one, that would be great. I'll continue on with the other tasks on the list for now, and start looking at all the open PRs.

Dreamsorcerer:
OK, I think as we sort out typing, aiojobs and other recent issues, it looks like there will need to be some significant changes to the design.

So, I'll push a release as soon as we have Python 3.11 sorted. The tests on 3.11 are failing, seemingly with the caches not retrieving values: #618.

Dreamsorcerer:
Right, looks like we've got 3.11 working (thanks @a68366). I'll look at doing a release on Friday.

If there's any more PRs people want merged, now is the time. There are quite a few reasonable looking PRs that people haven't bothered to add tests to. So, if anyone wants to finish up any of those, that'd be great.

Dreamsorcerer:
OK, finally out.

If anyone wants to contribute to the next major release, I've tagged several things to do under the 1.0 milestone which are likely to result in backwards-incompatible changes.
https://github.com/aio-libs/aiocache/milestone/8

pshafer-als:
@Dreamsorcerer  Thank you for driving all these updates and getting them released.

PR #546, opened 2021-08-21
[Bugfix] Add redis-py >= 4.2.0 support

laggardkernel:
2022.05.12: switched backend from aioredis` to `redis-py >= 4.2.0`.

---

**Update**: 

- To test with `aioredis` 2.0.0 automatically. Travis and tox configurations need to be updated. 
- Fixes aio-libs/aiocache#543

## What do these changes do?

aioredis 2.0.0 released with breaking API change. Update current backend implementation of `airedis/backends/redis.py`.

Breaking aioredis changes encounter during making this PR. aio-libs/aioredis-py#1109

Nope. Lower level redis backend change. No public API affected.

## Checklist

### DIFF
--- a/aiocache/backends/redis.py
+++ b/aiocache/backends/redis.py
@@ -103,13 +103,12 @@ class RedisBackend:
         return await self.client.setex(key, ttl, value)
 
     async def _cas(self, key, value, token, ttl=None, _conn=None):
-        args = [value, token]
+        args = [key, value, token]
         if ttl is not None:
             if isinstance(ttl, float):
                 args += ["PX", int(ttl * 1000)]
             else:
                 args += ["EX", ttl]
-        args = [key] + args
         res = await self._raw("eval", self.CAS_SCRIPT, 1, *args, _conn=_conn)
         return res
 
--- a/aiocache/backends/redis.py
+++ b/aiocache/backends/redis.py
@@ -48,7 +48,7 @@ class RedisBackend:
         super().__init__(**kwargs)
         if loop is not None:
             warnings.warn(
-                "Parameter 'loop' has been obsolete on aiocache >= 0.12.0",
+                "Parameter 'loop' has been deprecated since aiocache 0.12",
                 DeprecationWarning,
             )
         if pool_min_size is not _NOT_SET:
--- a/aiocache/backends/redis.py
+++ b/aiocache/backends/redis.py
@@ -72,7 +72,7 @@ class RedisBackend:
         # (decode_responses=False), because some of the values are saved as
         # bytes directly, like pickle serialized values, which may raise an
         # exception when decoded with 'utf-8'.
-        self.client = redis.Redis(host=self.endpoint, port=self.port, db=self.db
+        self.client = redis.Redis(host=self.endpoint, port=self.port, db=self.db,
                                   password=self.password, decode_responses=False,
                                   socket_connect_timeout=self.create_connection_timeout,
                                   max_connections=self.pool_max_size)
--- a/aiocache/backends/redis.py
+++ b/aiocache/backends/redis.py
@@ -65,10 +65,8 @@ class RedisBackend:
         #  but doesn't wait for connection reuse when number of conns
         #  exceeds the limit. While in aioredis 1.x, it waits asynchronously
         #  via asyncio.Condition. So max pool size defaults to None in redis.
-        if pool_max_size is None:
-            self.pool_max_size = None
-        else:
-            self.pool_max_size = int(pool_max_size)
+        # TODO: Remove int() call some time after adding type annotations.
+        self.pool_max_size = None if pool_max_size is None else int(pool_max_size)
         self.create_connection_timeout = (
             float(create_connection_timeout) if create_connection_timeout else None
         )
--- a/aiocache/backends/redis.py
+++ b/aiocache/backends/redis.py
@@ -134,13 +134,9 @@ class RedisBackend:
     async def __multi_set_ttl(self, flattened, ttl):
         async with self.client.pipeline(transaction=True) as p:
             p.execute_command("MSET", *flattened)
-            if isinstance(ttl, float):
-                ttl = int(ttl * 1000)
-                for key in flattened[::2]:
-                    p.pexpire(key, time=ttl)
-            else:
-                for key in flattened[::2]:
-                    p.expire(key, time=ttl)
+            ttl, exp = (int(ttl * 1000), p.pexpire) if isinstance(ttl, float) else (ttl, p.expire)
+            for key in flattened[::2]:
+                exp(key, time=ttl)
             await p.execute()
 
     async def _add(self, key, value, ttl=None, _conn=None):
--- a/aiocache/backends/redis.py
+++ b/aiocache/backends/redis.py
@@ -67,21 +67,15 @@ class RedisBackend:
             float(create_connection_timeout) if create_connection_timeout else None
         )
 
-        redis_kwargs = {
-            "host": self.endpoint,
-            "port": self.port,
-            "db": self.db,
-            "password": self.password,
-            # NOTE: decoding can't be controlled on API level after switching to
-            #  redis, we need to disable decoding on global/connection level.
-            #  Cause some of the values are save as bytes directly, like pickle
-            #  serialized values, which may raise exc when decoded with 'utf-8'.
-            "decode_responses": False,
-            # parameter names are different
-            "socket_connect_timeout": self.create_connection_timeout,
-            "max_connections": self.pool_max_size,
-        }
-        self.client = redis.Redis(**redis_kwargs)
+        # NOTE: decoding can't be controlled on API level after switching to
+        # redis, we need to disable decoding on global/connection level
+        # (decode_responses=False), because some of the values are saved as
+        # bytes directly, like pickle serialized values, which may raise an
+        # exception when decoded with 'utf-8'.
+        self.client = redis.Redis(host=self.endpoint, port=self.port, db=self.db
+                                  password=self.password, decode_responses=False,
+                                  socket_connect_timeout=self.create_connection_timeout,
+                                  max_connections=self.pool_max_size)
 
     async def _get(self, key, encoding="utf-8", _conn=None):
         value = await self.client.get(key)
--- a/aiocache/backends/redis.py
+++ b/aiocache/backends/redis.py
@@ -93,13 +93,10 @@ class RedisBackend:
         return await self._get(key, encoding=encoding, _conn=_conn)
 
     async def _multi_get(self, keys, encoding="utf-8", _conn=None):
-        values = []
-        for value in await self.client.mget(*keys):
-            if encoding is None or value is None:
-                values.append(value)
-            else:
-                values.append(value.decode(encoding))
-        return values
+        values = await self.client.mget(*keys)
+        if encoding is None:
+            return values
+        return [v if v is None else v.decode(encoding) for v in values]
 
     async def _set(self, key, value, ttl=None, _cas_token=None, _conn=None):
         if _cas_token is not None:
--- a/aiocache/backends/redis.py
+++ b/aiocache/backends/redis.py
@@ -1,8 +1,8 @@
 import itertools
 import warnings
 
-import aioredis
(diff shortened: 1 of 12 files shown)


gerardcl:
hi @laggardkernel the PR LGTM! I like it! 👍  let me know if you want me to help on anything missing from the list! thanks a ton

markzporter:
Would love to see this merged

JimNero009:
Is this project still maintained? How can we as a group go about getting this merged? For us, this is blocking the migration of some keen souls to use Python 3.10 in the company.

Boring-Mind:
> Is this project still maintained? How can we as a group go about getting this merged? For us, this is blocking the migration of some keen souls to use Python 3.10 in the company.

You can create a fork from this project and make any adjustments you want. It's pretty easy. If you need to download the package from pip, please upload it to pypi.

JimNero009:
We may indeed have to go down that route if there is no chance of this change getting into the codebase any time soon -- is this the case? The changes here seem functionally complete? 

In other words, is it 'official' that this codebase is no longer maintained?

argaen:
Hi, sadly I'm not maintaining this project anymore but I'm happy to invite people to the project in case anyone wants to volunteer.

pboers1988:
@argaen has anybody volunteerd as of yet?

Dreamsorcerer:
If anyone wants to make a start, feel free to start PRs for migrating the CI to Github. See: https://github.com/aio-libs/aiocache/issues/536#issuecomment-1024448383

Dreamsorcerer:
I think keeping up the multi-version support is unsustainable and has little benefit. I'd like to just set the minimum version to `aioredis>=2` for the next release and keep things simple. If someone has time to update this PR to only use v2, that would be great. Otherwise, I'll look at updating later.

laggardkernel:
@Dreamsorcerer Okay, I'd like to have a try to adapt the PR with only `aioredis>=2` support when I have spare time.

Mark90:
@laggardkernel Did you get blocked on something, or just didn't get around to this? Otherwise, me (or someone else) could try to pick this up.

Dreamsorcerer:
Yes, might as well move to that library then. Though it's the same work, it looks like they've merged aioredis upstream, so the only difference currently is the import statement.

laggardkernel:
WIP: Updated the PR with only aioredis 2.x support. The test is fixed and passed, but the pytest fixtures may need some updates. I'll check that later.

> Would you agree it makes more sense to refactor aiocache to use redis-py than to make it ready for aioredis-py 2.x when the latter is likely to be deprecated?

I have read both `aioredis-py` and `redis-py` before. The last version of `redis-py` I read is still 4.1. I'll check the latest release of redis-py soon.

---

@Mark90 The time when I opened the PR I was unemployed. I had a good time browsing and reading all kinds of Python libraries, learning how they are designed. Now i'm back to work and kind of busy. Sorry. 🙇‍♂️

laggardkernel:
aioredis 2.x only support is done.

This PR is firstly made with backward compatiblity taken into consideration.
It supported for both aioredis 1.x and 2.x.
After the author agreed to drop support for aioredis 1.x, I went into a wrong
direction that kept the old design of redis cache backend, which managed
conn and pool manually. This is required when integrating with aioredis 1.x
but not with 2.x

Luckily, I realised it after a while, and refactored the code to be similar
with the memecached cache backend. It's much simpler and much more elegant now.

---

Most of my time is not spent on refactoring the redis cache backend code.
It's taken to make a correct mock of the `aioredis.Redis` instance.
FYI, methods like `.get()` `.set()` on `Redis` are not coroutine. In fact,
when we do `await redis.get()` we're await its returned value, which is a coroutine.

Autospeccing `aioredis.Redis` doesn't exactly reflect the above behavior.
I mocked the `.get()`, `.set()` methods with `AsyncMock` now to pass the tests.
Is there a better way to mock the methods more precisely?
I guess it doesn't matter, cause all we wanna do is to ensure `Redis.method` gets called.
Whether `Redis.method` communicates with the redis server correctly is tested in `aioredis`.

Sorry, I haven't had time to check async support in `redis-py` yet.

laggardkernel:
Done. 

The `redis >= 4.2.0` async support is based on `aioredis` and the APIs are compatible. I switched to from `aioredis` to `redis` easily. All redis related tests are fixed and passed on my local machine.

There're some breaking changes compared with `aioredis 1.x`, I'll detail them as code comment conversions.

laggardkernel:
Sorry, I thought the pre-commit hooks was enough to detect the linting err. Maybe some of my commits were made without git hooks installed yet.

Anyway, I fixed the linting err with `make lint`. I re-disabled the performance tests. Haven't dig it deeper to fix them yet. Not only redis, there're also some problems with the memcached performance tests as well, during the test running on my local machine.

BTW, I rebased the PR after the "test key refactoring" merged today.

joeblackwaslike:
@laggardkernel @argaen Do you guys need someone to take this PR over the finish line?  If you want to add me as a maintainer I can finish any remaining tasks necessary and cut a new release.  Thx

laggardkernel:
The PR is completed. Just need another run of the test by the CI, which is not triggered automatically. Part of the performance test is disabled, cause it doesn't fit the connection pool in `redis-py`.

@joeblackwaslike Glad about the help you offered. I'm afraid there's nothing we could do. Just the project lacks maintenance. If you need to use aioredis in your own project, you'd better fork the repo, merge this PR and build a package by yourself.

argaen:
Hi, as mentioned above, I'm not maintaining this project anymore but I'm happy to invite any maintainers who would like to contribute. @laggardkernel @joeblackwaslike let me know and I'll invite you

Krukov:
Hi there,

If someone here looking for alternative with more up to date support:

https://github.com/Krukov/cashews  - ready to be used and contribute
 - using redis-py
 - python 3.7-3.11

Dreamsorcerer:
I'll refresh myself on where this is at in a couple of days.
Tests probably just need to have master merged in order to run, maybe a force-push messed it up or something.

aiocache/backends/redis.py
```diff
@@ -52,163 +39,171 @@ def __init__(
...
-        self.pool_min_size = int(pool_min_size)
-        self.pool_max_size = int(pool_max_size)
+        # NOTE: In redis and aioredis 2.x, ConnectionPool raises ConnectionError
+        #  but doesn't wait for connection reuse when number of conns
+        #  exceeds the limit. While in aioredis 1.x, it waits asynchronously
+        #  via asyncio.Condition. So max pool size defaults to None in redis.
+        if pool_max_size is None:
+            self.pool_max_size = None
+        else:
+            self.pool_max_size = int(pool_max_size)
```
Dreamsorcerer, line 71:
```suggestion
        self.pool_max_size = None if pool_max_size is None else int(pool_max_size)
```

Dreamsorcerer: We should probably just look at adding type annotations and then not doing this `int()` conversion...

aiocache/backends/redis.py
```diff
@@ -52,163 +39,171 @@ def __init__(
...
         self.endpoint = endpoint
         self.port = int(port)
         self.db = int(db)
         self.password = password
-        self.pool_min_size = int(pool_min_size)
-        self.pool_max_size = int(pool_max_size)
+        # NOTE: In redis and aioredis 2.x, ConnectionPool raises ConnectionError
+        #  but doesn't wait for connection reuse when number of conns
+        #  exceeds the limit. While in aioredis 1.x, it waits asynchronously
+        #  via asyncio.Condition. So max pool size defaults to None in redis.
```
Dreamsorcerer, line 67:
```suggestion
```

Nice to know why the change in the PR, but probably doesn't need to be committed to the code.

aiocache/backends/redis.py
```diff
@@ -52,163 +39,171 @@ def __init__(
...
+            # NOTE: decoding can't be controlled on API level after switching to
+            #  redis, we need to disable decoding on global/connection level.
+            #  Cause some of the values are save as bytes directly, like pickle
+            #  serialized values, which may raise exc when decoded with 'utf-8'.
+            "decode_responses": False,
+            # parameter names are different
+            "socket_connect_timeout": self.create_connection_timeout,
+            "max_connections": self.pool_max_size,
+        }
+        self.client = redis.Redis(**redis_kwargs)
```
Dreamsorcerer, line 90: Better to put the kwargs in directly. You lose typing information with a dict, so mypy is unlikely to catch wrong types getting passed in (e.g. if upstream API changes).

aiocache/backends/redis.py
```diff
@@ -52,163 +39,171 @@ def __init__(
...
-    @conn
     async def _cas(self, key, value, token, ttl=None, _conn=None):
         args = [value, token]
         if ttl is not None:
             if isinstance(ttl, float):
                 args += ["PX", int(ttl * 1000)]
             else:
                 args += ["EX", ttl]
-        res = await self._raw("eval", self.CAS_SCRIPT, [key], args, _conn=_conn)
+        args = [key] + args
```
Dreamsorcerer, line 127: Why not just add key to the start of the list on line 121?

aiocache/backends/redis.py
```diff
@@ -52,163 +39,171 @@ def __init__(
...
+        return await self.client.setex(key, ttl, value)
 
-    @conn
     async def _cas(self, key, value, token, ttl=None, _conn=None):
         args = [value, token]
         if ttl is not None:
             if isinstance(ttl, float):
                 args += ["PX", int(ttl * 1000)]
             else:
                 args += ["EX", ttl]
```
Dreamsorcerer, line 126: On an unrelated point, these should probably be `args.extend(...)`, rather than doing concatenation.

tests/performance/conftest.py
```diff
@@ -1,21 +1,26 @@
...
-# TODO: Update aioredis and fix tests.
+
+# TODO:
+#  1. concurrency_error_rates test doesn't work as expected with redis-py,
+#  where both ConnectionPool and BlockingConnectionPool raise ConnectionError
+#  but don't wait for connection reuse when number of conns exceeds the limit.
+#  That's why pool in redis-py uses an unlimit number connections by default.
+#  So no HTTP req should fails in the concurrency_error_rates test with redis-py.
+#  2. On my local machine, test_memcached_getsetdel() fails, it doesn't reach
+#  the performance target.
```
Dreamsorcerer, line 13: Let's move this into an issue if we're not fixing it immediately.

tests/ut/backends/test_redis.py
```diff
@@ -1,319 +1,268 @@
...
-
 
 @pytest.fixture
-def redis_pool(redis_connection):
-    class FakePool:
-        def __await__(self):
-            yield
-            return redis_connection
+def redis_pipeline():
+    pipeline = AsyncMock(name="redis_pipeline")
```
Dreamsorcerer, line 14: Can you try with `create_autospec()`? Code would be more robust, and the next few lines can probably be removed as well.

tests/ut/backends/test_redis.py
```diff
@@ -1,319 +1,268 @@
...
-        )
-
-    @pytest.mark.asyncio
-    async def test_get(self, redis, redis_connection):
-        await redis._get(Keys.KEY)
-        redis_connection.get.assert_called_with(Keys.KEY, encoding="utf-8")
+    async def test_get(self, redis):
+        redis.client.get.return_value = b"value"
+        assert await redis._get(Keys.KEY) == "value"
+        redis.client.get.assert_called_with(Keys.KEY)
```
Dreamsorcerer, line 108: The original test use `.assert_called_with(Keys.KEY, encoding="utf-8")`?

tests/ut/backends/test_redis.py
Dreamsorcerer: This is also on my list of tasks (https://github.com/aio-libs/aiocache/issues/536#issuecomment-1068390586), so can be looked at separately if you've not got the time.

Dreamsorcerer:
OK, looking good. Just some minor niggles I'd like resolved.
Sorry it's taken so long.

tests/ut/backends/test_redis.py
Dreamsorcerer: Because encoding is done in aiocache now.

PR #551, opened 2021-10-31
Use self instead of SimpleMemoryBackend

and-semakin:
## What do these changes do?

Use `self` instead of direct links to `SimpleMemoryBackend` class. This PR doesn't affect any functionality but allows users of the library inherit `SimpleMemoryBackend` and change its behaviour. Otherwise, changing the behavior is really difficult to do.

No.

## Checklist

- [x] Add a new news fragment into the `CHANGES` folder

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # CHANGELOG
 
+* Use `self` instead of hardcoded `SimpleMemoryBackend` to refer to cache and handlers [#551](https://github.com/aio-libs/aiocache/pull/551) - Andrey Semakin
 
 ## 0.11.1 (2019-07-31)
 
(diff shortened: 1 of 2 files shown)


Fogapod:
This is not a correct fix since you still use global cache and handlers. They are not set on init

and-semakin:
@Fogapod exactly, I didn’t even try to fix the issue with global cache state. My only purpose here was just to make inheritance easier. It’s just one little step in the correct direction.

Dreamsorcerer:
Looks good, sorry for the delay. Could you update with latest master and ensure the tests are all passing. Thanks.

Dreamsorcerer:
I think this is solved now with #562. Please reopen if not.

Issue #554, opened 2022-01-21
aiomcache 0.7.0 breaks aiocache[memcached]


minorgod:
New release 0.7.0 of aiomcache breaks aiocache[memcached]. 

https://github.com/aio-libs/aiocache/blob/9c8b07fe759990dcb2d4d5f4e40d13d2cc36d58f/setup.py#L43

Version range probably needs to be changed to something like:
aiomcache>=0.5.2,< 0.7.0

minorgod:
Also, FYI,  I assume this is not related to the Python version, but this was discovered while running in Python 3.8 Docker image. Haven't tested in Python 3.9, but restricting the aiomcache package to version to < 0.7.0 does fix the issue for me in Python 3.8.

Dreamsorcerer:
Working fine in master now, shouldn't be a problem in 0.12.

PR #555, opened 2022-01-21
Restrict aiomcache to versions < 0.7.0

minorgod:
aiomcache version 0.7.0 breaks aiocache[memcached] in Python 3.8 and probably others:
When trying to create the default cache, aiocache throws:
```
TypeError: __init__() got an unexpected keyword argument 'loop'
```
Restricting aimcache lib < 0.7.0 fixes the issue. 

## Related issue number

https://github.com/aio-libs/aiocache/issues/554


Dreamsorcerer:
Would be better to update the code to be compatible, and drop any deprecated loop parameters so everything continues working on Python 3.10+.

Dreamsorcerer:
Latest code works, so this isn't relevant anymore.

Issue #531, opened 2021-02-04
One dict for any SimpleMemoryBackend instances. Why is that?


AntonBezdenezhnykh:
Hello, I'm trying to use several memory caches for different functions. And I found that SimpleMemoryBackend uses only one dict for all instances. I'm curious why is that? What are underwater stones there?
It looks easy to start using separate dicts for each SimpleMemoryBackend instance.

and-semakin:
Agree. This behavior is a little bit confusing, especially in tests when you receive some unexpected data in the cache that was put there by a previous test. It becomes difficult to write tests that do some assertions about cache.

PR #562, opened 2022-03-18
make SimpleMemoryBackend store state in instance instead of class

Fogapod:
## What do these changes do?

Make SimpleMemoryBackend store state in instance instead of class.
While current behaviour matches other caches (same connection args == same cache; zero connection args in memory case), current design seems bad since it does not allow extending and customizing memory cache in future.

Tests adjusted accordingly.

Yes, if user for some reason expected different SimpleMemoryBackend objects to have shared storage. This is probably a breaking change for some weird setups.

## Related issue number

Resolves: #531

Resolves: #479
There is a PR aimed to fix it: https://github.com/aio-libs/aiocache/pull/523
But using namespace seem to not be sufficient. Instance local cache must be used for consistent behaviour.

## Checklist

### DIFF
--- a/aiocache/backends/memory.py
+++ b/aiocache/backends/memory.py
@@ -1,5 +1,5 @@
 import asyncio
-from typing import Dict
+from typing import Any, Dict
 
 from aiocache.base import BaseCache
 from aiocache.serializers import NullSerializer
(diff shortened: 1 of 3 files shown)


aiocache/backends/memory.py
```diff
@@ -10,32 +10,32 @@ class SimpleMemoryBackend:
     Wrapper around dict operations to use it as a cache backend
     """
 
-    _cache: Dict[str, object] = {}
-    _handlers: Dict[str, asyncio.TimerHandle] = {}
-
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
 
+        self._cache: Dict[str, Any] = {}
```
Dreamsorcerer, line 16:
```suggestion
        self._cache: Dict[str, object] = {}
```

I think our internal code should be safe regardless of the value.

Dreamsorcerer (commented):
Looks good, sorry for the delay.

Could you update to latest master?

aiocache/backends/memory.py
Fogapod: Not sure what safety means here. I haven't seen `objeft` used this way anywhere since `Any` type is spicifically made for cases where you don't care about type

Fogapod:
Resolved conflicts

aiocache/backends/memory.py
Dreamsorcerer: No, `Any` is used to disable type checking because you have no way to guarantee type safety. `object` is type safe, meaning that the code is expected to work with anything passed to it.

For example, `print()` works with anything: https://github.com/python/typeshed/blob/main/stdlib/builtins.pyi#L1511
It is always type safe.

Dreamsorcerer: An improvement that we could look at later, would be to make these classes `Generic`, then we can make the value `_T` and have the methods `.get()` etc. return `_T`. But, that's something for another PR.

tests/ut/backends/test_memory.py
```diff
@@ -6,21 +6,21 @@
 from aiocache.backends.memory import SimpleMemoryBackend, SimpleMemoryCache
 from aiocache.base import BaseCache
 from aiocache.serializers import NullSerializer
+
```
Dreamsorcerer, line 9:
```suggestion
```

Dreamsorcerer:
Should probably have a regression test too, but I'll merge this in for now, as it's pretty straightforward.

Dreamsorcerer:
Ah, seems you missed updating some tests. 28 failing tests currently.

Fogapod:
I'll fix those and test locally a bit later

Fogapod:
These weren't tests, there was one more line using class var added after this PR, now fixed

Dreamsorcerer:
Ah, I clearly didn't look closely enough.

Fogapod:
I followed CONTRIBUTING.rst and ran `make test` after installing dev dependencies. All my local tests except memcached completed successfully (i don't have it installed).
Is there something I'm doing wrong

Dreamsorcerer:
I've not looked at those instructions, so they could be outdated. Ultimately, whatever the CI does, is the correct way to run it: https://github.com/aio-libs/aiocache/blob/master/.github/workflows/ci.yml#L103-L106

Dreamsorcerer:
It's just 1 test class failing now, so should be able to run it with `pytest tests/acceptance/test_plugins.py::TestHitMissRatioPlugin`

Dreamsorcerer:
Looks like it just needs this setup line needs to be updated: `SimpleMemoryBackend._cache = data` in both tests.

Fogapod:
`tests/acceptance/test_plugins.py::TestHitMissRatioPlugin` now passes locally and there are no more uses of `SimpleMemoryBackend` in code

Issue #563, opened 2022-04-11
multi_cached not preserving original keys when calling the underlying function


haizhou:
@jkgenser and I were testing `multi_cached` in our project, and found that transformed keys are being passed into the underlying function call after `key_builder` is called.

Dreamsorcerer:
Needs a test/reproducer.

pshafer-als:
I don't want to disrupt PR #564, so I'm moving this bit of discussion here instead.

> If I'm understanding correctly, you're suggesting in interpretation 2, that a function can return keys which don't match the input? Like:
>
> ```
> def foo(keys=(1, 2)):
>    return {"a": "foo", "b": "bar"}
> ```

Well, that's an extreme example, but I suppose "legal" under[ Interpretation 2](https://github.com/aio-libs/aiocache/pull/564#issuecomment-1380574858). :) 

I was expecting something more like this...

```
def sales(mail_codes=(10101, 90210)):
    return {"sales_10101": 546_295, "sales_90210": 133_420}

def financials(year=(2021, 2022, 2023), stats=("revenue", "expenses")):
    return {"2021.revenue": 245_364, "2021.expenses": 546_295, "2022.revenue": 1_216_892, "2022.expenses": 748_398, "2023.revenue": 83_294, "2023.expenses": 56_492}
```

I agree that these functions should know nothing about the cache.

My interpretation was that:
1. `key_builder(key, func, *args, **kwargs)` in `@multicached(..., key_builder=...)` converts the input keys from the decorated function into the keys that are returned by that function.
2. The cache used by `multicached` calls its `build_key(key, namespace)` member to optionally add a namespace to the keys returned by the decorated function to avoid a key naming clash.

---
On subsequent review, I'm realizing why this is confusing.

In decorator, before _**reading**_ from the cache, the cache keys are built by `get_cache_keys()`, which calls `key_builder()` on the keys referenced by `keys_from_attr`.  The keys returned by the decorated function are not used here.

https://github.com/aio-libs/aiocache/blob/5007dc25278877a47b94a4dbe43a4afec27951ac/aiocache/decorators.py#L377-L380

When `set_in_cache()` _**writes**_ to the cache, it calls `key_builder()` on the keys returned by the decorated function. `keys_from_attr` and `get_cache_keys()` are not used here.

https://github.com/aio-libs/aiocache/blob/79c3f7008a919b33e7aa63eab85a7d99c65a3040/aiocache/decorators.py#L355-L360

[...]

PR #564, opened 2022-04-11
FIX: preserve original keys after key_builder transformation

haizhou:
## What do these changes do?

Preserve original keys with the underlying function call after `key_builder` transformation takes place in `multi_cached`

There shouldn't be if the intention is to allow differently formatted cache keys than the original keys passed in.

## Related issue number

https://github.com/aio-libs/aiocache/issues/563

## Checklist


Dreamsorcerer:
Looks right to me, but can you add a test to demonstrate the problem (and merge master).

Sorry for the delay.

aiocache/decorators.py
```diff
@@ -310,19 +309,19 @@ async def decorator(
...
-                    missing_keys.append(key)
+                    missing_keys.append(orig_key)
                 else:
-                    partial[key] = value
+                    partial[orig_key] = value
             if values and None not in values:
                 return partial
         else:
-            missing_keys = list(keys)
+            missing_keys = list(orig_keys)
```
pshafer-als, line 324: I think you're right that the keys from `keys_from_attr` should get stored in `missing keys`, because those get passed to the callable.

aiocache/decorators.py
```diff
@@ -310,19 +309,19 @@ async def decorator(
...
-            values = await self.get_from_cache(*keys)
-            for key, value in zip(keys, values):
+            values = await self.get_from_cache(*cache_keys)
+            for orig_key, value in zip(orig_keys, values):
                 if value is None:
-                    missing_keys.append(key)
+                    missing_keys.append(orig_key)
                 else:
-                    partial[key] = value
+                    partial[orig_key] = value
```
pshafer-als, line 320: No, I think that `partial` should use the mapped keys returned by `key_builder`. These keys need to match the keys in the returned data, not the keys passed into the callable.

aiocache/decorators.py
```diff
@@ -310,19 +309,19 @@ async def decorator(
...
+        orig_keys, cache_keys, new_args, args_index = self.get_cache_keys(f, args, kwargs)
 
         if cache_read:
-            values = await self.get_from_cache(*keys)
-            for key, value in zip(keys, values):
+            values = await self.get_from_cache(*cache_keys)
+            for orig_key, value in zip(orig_keys, values):
                 if value is None:
-                    missing_keys.append(key)
+                    missing_keys.append(orig_key)
```
pshafer-als, line 318: Agreed.  See the comment below regarding `missing_keys`.

aiocache/decorators.py
```diff
@@ -310,19 +309,19 @@ async def decorator(
     ):
         missing_keys = []
         partial = {}
-        keys, new_args, args_index = self.get_cache_keys(f, args, kwargs)
+        orig_keys, cache_keys, new_args, args_index = self.get_cache_keys(f, args, kwargs)
```
pshafer-als, line 312: Consider naming these as `in_keys` and `out_keys` for clarity.  Respectively they are the keys that are passed in to the callable, and the keys that are returned by the callable (`out_keys` are also the translated keys returned by `key_builder`).

If these `in_keys` are only used internally, then it might make sense to leave `get_cached_keys()` unmodified (i.e., revert it).  Then instead `in_keys` could be generated here from `_get_args_dict()`. `get_cache_keys()` is a public member function.

aiocache/decorators.py
pshafer-als: > No, I think that `partial` should use the mapped keys returned by `key_builder`. These keys need to match the keys in the returned data, not the keys passed into the callable.

This is under the assumption of [Interpretation 2](https://github.com/aio-libs/aiocache/pull/564#issuecomment-1380574858).

aiocache/decorators.py
pshafer-als: > Consider naming these as `in_keys` and `out_keys` for clarity. Respectively they are the keys that are passed in to the callable, and the keys that are returned by the callable (`out_keys` are also the translated keys returned by `key_builder`).

This is under the assumption of [Interpretation 2](https://github.com/aio-libs/aiocache/pull/564#issuecomment-1380574858).

haizhou:
Agree with @pshafer-als on need for clarification about intended use case for `key_builder` in `multi_cached`. I added an example, but as of the latest release, it is raising a KeyError on https://github.com/aio-libs/aiocache/blob/0.11.1/aiocache/decorators.py#L366 because the key returned from the result has the already transformed key by `key_builder` but is passed through again

Dreamsorcerer:
> Hmmm, there still seems to be some ambiguity regarding the dict keys that are returned by the decorated function and the role of `key_builder` in `multi_cached`. I see two interpretations. Can someone help to clarify the intended usage (or the prominent use case)?

If I'm understanding correctly, you're suggesting in interpretation 2, that a function can return keys which don't match the input? Like:
```
def foo(keys=(1, 2)):
    return {"a": "foo", "b": "bar"}
```

That's not my understanding of the documentation.

> By default, the callable’s name and call signature are not incorporated into the cache key, so if there is another cached function returning a dict with same keys, those keys will be overwritten. To avoid this, use a specific namespace in each cache decorator or pass a key_builder.

So, the primary purpose is to avoid cache key collisions with other functions. Not so sure why you'd use it over namespace though...

> the values of keys_from_attr will be transformed before requesting them from the cache.

So, we use key_builder for the key to request from cache.

> Equivalently, the keys in the dict-like mapping returned by the decorated callable will be transformed before storing them in the cache.

And we use key_builder on the returned keys to set them in the cache.

In other words, it seems like key_builder just transforms keys to deal with the cache. The function should have no knowledge of the caching details, and the transformed keys should not be visible outside of the caching code.

Which means something like this should work:
```
def my_func(*args):
    ...

f1 = multi_cached("args", key_builder=builder1)(my_func)
f2 = multi_cached("args", key_builder=builder2)(my_func)
my_func(1)
f1(1)
f2(1)
# All 3 calls should return the same result, with different caches used for the latter 2 functions.
```

Dreamsorcerer:
So, I think the changes here are still correct. If we can get a new test for this (in tests/ut/test_decorators.py using a mock cache should work), then we can merge this.

tests/ut/test_decorators.py
```diff
@@ -590,6 +590,13 @@ async def bar():
 
         assert foo.cache != bar.cache
 
+    async def test_key_builder(self):
+        @multi_cached("keys", key_builder=lambda key, _, keys: key + 1)
+        async def incr(keys=None):
+            return {k: k + 1 for k in keys}
+        
+        assert incr(keys=[1]) == {1: 3}
```
Dreamsorcerer, line 598:
```suggestion
        @multi_cached("keys", key_builder=lambda key, _, keys: key + 1)
        async def incr(keys=None):
            return {k: k * 3 for k in keys}
        
        assert incr(keys=(1,)) == {1: 3}
        cached_value = await incr.cache.get(2)
        assert cached_value == 3
        assert not await incr.cache.exists(1)
```

Maybe something like this is more comprehensive.

examples/multicached_decorator.py
```diff
@@ -39,6 +49,8 @@ async def test_multi_cached():
     await cache.delete("d")
     await cache.close()
 
+    await multi_cached_key_builders(keys=(1, 2)) == {1: "Z", 2: "Y"}
```
Dreamsorcerer, line 52: I'm not sure we need more in the example, so I'd suggest reverting the changes to this file (the code is wrong currently anyway).

tests/ut/test_decorators.py
```diff
@@ -590,6 +590,16 @@ async def bar():
...
 
+    async def test_key_builder(self):
+        @multi_cached("keys", key_builder=lambda key, _, keys: key + 1)
+        async def incr(keys=None):
+            return {k: k * 3 for k in keys}
+        
+        assert await incr(keys=(1,)) == {1: 3}
+        cached_value = await incr.cache.get(2)
+        assert cached_value == 3
+        assert not await incr.cache.exists(1)
```
Dreamsorcerer, line 601:
```suggestion
        @multi_cached("keys", key_builder=lambda key, _, keys: key + 1)
        async def f(keys=None):
            return {k: k * 3 for k in keys}


        assert await f(keys=(1,)) == {1: 3}
        cached_value = await f.cache.get(2)
        assert cached_value == 3
        assert not await f.cache.exists(1)
```

Dreamsorcerer: Huh, the commit button is not there...

Dreamsorcerer: Oh, you must not have allowed maintainers write access...

Dreamsorcerer:
@haizhou There are 6 failing tests, due to the change of return value in .get_cache_keys(). Can you enable write access to you branch or update those tests?

All the failing tests are under tests/ut/test_decorators.py::TestMultiCached

tests/ut/test_decorators.py
```diff
@@ -590,6 +590,17 @@ async def bar():
 
         assert foo.cache != bar.cache
 
+    async def test_key_builder(self):
+        @multi_cached("keys", key_builder=lambda key, _, keys: key + 1)
+        async def f(keys=None):
+            return {k: k * 3 for k in keys}
+
+
```
Dreamsorcerer, line 598:
```suggestion
```

I also misread the flake8 error.

haizhou:
> @haizhou There are 6 failing tests, due to the change of return value in .get_cache_keys(). Can you enable write access to your branch or update those tests?
> 
> All the failing tests are under tests/ut/test_decorators.py::TestMultiCached

@Dreamsorcerer thanks for the collaboration. I'm not sure if we have enough seats left for adding an outside collaborator to our repos, but I took an attempt at fixing the tests

Dreamsorcerer:
> > @haizhou There are 6 failing tests, due to the change of return value in .get_cache_keys(). Can you enable write access to your branch or update those tests?
> > All the failing tests are under tests/ut/test_decorators.py::TestMultiCached
> 
> @Dreamsorcerer thanks for the collaboration. I'm not sure if we have enough seats left for adding an outside collaborator to our repos, but I made an attempt to fix the tests

No problem. I think there's usually a checkbox to enable maintainer write access when opening a PR (not sure how to find it after opening though). That shouldn't affect those seats (and probably only applies to the branch, rather than the entire repo).

tests/ut/test_decorators.py
```diff
@@ -395,26 +395,27 @@ def test_alias_takes_precedence(self, mock_cache):
...
     def test_get_cache_keys_missing_kwarg(self, decorator):
-        assert decorator.get_cache_keys(stub_dict, (), {}) == ([], [], -1)
+        assert decorator.get_cache_keys(stub_dict, (), {}) == ([], [], [], -1)
 
     def test_get_cache_keys_arg_key_from_attr(self, decorator):
         def fake(keys, a=1, b=2):
             """Dummy function."""
 
-        assert decorator.get_cache_keys(fake, (["a"]), {}) == (["a"], [["a"]], 0)
+        assert decorator.get_cache_keys(fake, (["a"]), {}) == ("a", ["a"], ["a"], 0)
```
Dreamsorcerer, line 410:
```suggestion
        assert decorator.get_cache_keys(fake, (["a"],), {}) == ("a", ["a"], ["a"], 0)
```

I think the expected output was correct before, someone just clearly messed up the input. The test passed before by luck, because `["a"][0] == "a"[0]`.

Issue #542, opened 2021-06-11
Redis cache: `get()` does not work for incremented values (returns '_pickle.UnpicklingError: could not find MARK')


AIGeneratedUsername:
Steps to reproduce:

    await cache.increment("foobareggs", delta=10)
    assert await cache.get("foobareggs")

Result:
```
>       assert await cache.get("foobareggs")

test_common_cache.py:40: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../src/common/cache.py:230: in get
    return await self.cache.get(key, default, loads_fn=loads_fn, namespace=namespace or self.namespace, timeout=timeout or self.timeout, **kwargs)
../../venv/lib/python3.9/site-packages/aiocache/base.py:61: in _enabled
    return await func(*args, **kwargs)
../../venv/lib/python3.9/site-packages/aiocache/base.py:45: in _timeout
    return await asyncio.wait_for(func(self, *args, **kwargs), timeout)
/usr/lib/python3.9/asyncio/tasks.py:481: in wait_for
    return fut.result()
../../venv/lib/python3.9/site-packages/aiocache/base.py:75: in _plugins
    ret = await func(self, *args, **kwargs)
../../venv/lib/python3.9/site-packages/aiocache/base.py:192: in get
    value = loads(await self._get(ns_key, encoding=self.serializer.encoding, _conn=_conn))
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <aiocache.serializers.serializers.PickleSerializer object at 0x7f5042157160>
value = b'10'

    def loads(self, value):
        """
        Deserialize value using ``pickle.loads``.
    
        :param value: bytes
        :returns: obj
        """
        if value is None:
            return None
>       return pickle.loads(value)
E       _pickle.UnpicklingError: could not find MARK

../../venv/lib/python3.9/site-packages/aiocache/serializers/serializers.py:119: UnpicklingError
```

citramon:
In this case it doesn't work with JsonSerializer too:
```
from aiocache.serializers import JsonSerializer
cache = Cache(Cache.MEMORY, serializer=JsonSerializer())
loop.run_until_complete(cache.set("a", 1))
True
[...]

PR #565, opened 2022-04-11
Hotfix increment

citramon:
## What do these changes do?

The problem is that increment function doesn't use serialize method.
Without it increment function works with several serializes only.

nope

## Related issue number

Resolves: [542](https://github.com/aio-libs/aiocache/issues/542)

## Checklist

- [X] Unit tests for the changes exist
- [X] Documentation reflects the changes
- [X] Add a new news fragment into the `CHANGES` folder

-----
[View rendered CHANGELOG.md](https://github.com/citramon/aiocache/blob/hotfix_increment/CHANGELOG.md)

### DIFF
--- a/aiocache/backends/redis.py
+++ b/aiocache/backends/redis.py
@@ -100,9 +100,11 @@ class RedisBackend:
         return await _conn.mget(*keys, encoding=encoding)
 
     @conn
-    async def _set(self, key, value, ttl=None, _cas_token=None, _conn=None):
+    async def _set(self, key, value, ttl=None, _cas_token=None, _conn=None, _keep_ttl=None):
         if _cas_token is not None:
             return await self._cas(key, value, _cas_token, ttl=ttl, _conn=_conn)
+        if _keep_ttl:
+            return await _conn.execute("SET", key, value, "KEEPTTL")
         if ttl is None:
             return await _conn.set(key, value)
         return await _conn.setex(key, ttl, value)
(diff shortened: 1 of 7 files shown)


aiocache/backends/redis.py
```diff
@@ -100,9 +100,11 @@ async def _multi_get(self, keys, encoding="utf-8", _conn=None):
         return await _conn.mget(*keys, encoding=encoding)
 
     @conn
-    async def _set(self, key, value, ttl=None, _cas_token=None, _conn=None):
+    async def _set(self, key, value, ttl=None, _cas_token=None, _conn=None, _keep_ttl=None):
         if _cas_token is not None:
             return await self._cas(key, value, _cas_token, ttl=ttl, _conn=_conn)
+        if _keep_ttl:
+            return await _conn.execute("SET", key, value, "KEEPTTL")
         if ttl is None:
```
Dreamsorcerer, line 108: Shouldn't this become `ellif`?

Dreamsorcerer:
I'm not sure this makes sense. It seems to me like `increment()` is meant to be an efficient, atomic operation on the backend. If the value is serialised to anything other than an `int` then this is obviously not possible.

If you're happy without the efficient, atomic part, then it seems like you may as well not use `increment()` and instead just do `set(k, get(k) + 1)`.

Dreamsorcerer:
Let me know if you want to work on this again, but this doesn't look like the right solution in the current form.

Issue #473, opened 2019-12-24
Deprecate using cache specific constructors


argaen:
`Cache` class was introduced as a proxy for instantiating the different backends. There should be only one way of doing things and `Cache` is the preferred way of doing that so will deprecate the others

pshafer-als:
This should probably also apply to creating the cache inside the decorator `@cached`.  It looks like this might have been overlooked.

https://github.com/aio-libs/aiocache/blob/02782c238aa8a6ca238ef938b3c70e017dc5a2ea/aiocache/decorators.py#L207-L208

PR #568 was submitted.

Dreamsorcerer:
This is an open issue, so doesn't look like anybody has deprecated anything yet.

Dreamsorcerer:
I'm not sure how well this'll play with static typing, so I suspect we'll have to change this again as we introduce typing to the library.

It seems slightly odd to me that we instantiate a wrapper class and pass the implementation class as an argument, rather than just using it directly. i.e. Why do `Cache(Cache.MEMORY)` and not just `Cache.MEMORY()` or `SimpleMemoryCache()`?

PR #568, opened 2022-04-24
Decorator @cached creates cache with the Cache() factory

pshafer-als:
* See #473 ~~: Deprecate using cache specific constructors~~
  ~~* https://github.com/aio-libs/aiocache/issues/473~~
  * `Cache` class was introduced as a proxy for instantiating the different
    backends. There should be only one way of doing things and `Cache` is
    the preferred way of doing that so will deprecate the others.
* This is now applied to `cached._get_cache()`

## What do these changes do?

Decorator `@cached` creates cache with the `Cache()` factory constructor, 
instead of the cache-specific constructor.

This should be invisible to the end user.

## Related issue number

#473: Deprecate using cache specific constructors

## Checklist

### DIFF
--- a/aiocache/decorators.py
+++ b/aiocache/decorators.py
@@ -205,12 +205,7 @@ class cached_stampede(cached):
 
 
 def _get_cache(cache=Cache.MEMORY, serializer=None, plugins=None, **cache_kwargs):
-    return Cache(
-        cache, 
-        serializer=serializer, 
-        plugins=plugins, 
-        **cache_kwargs,
-    )
+    return Cache(cache,  serializer=serializer,  plugins=plugins,  **cache_kwargs)
 
 
 def _get_args_dict(func, args, kwargs):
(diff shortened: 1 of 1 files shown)


pshafer-als:
Hi, do you have an feedback for this PR?

aiocache/decorators.py
```diff
@@ -205,7 +205,12 @@ async def decorator(self, f, *args, **kwargs):
 
 
 def _get_cache(cache=Cache.MEMORY, serializer=None, plugins=None, **cache_kwargs):
-    return cache(serializer=serializer, plugins=plugins, **cache_kwargs)
+    return Cache(
+        cache, 
+        serializer=serializer, 
+        plugins=plugins, 
+        **cache_kwargs,
+    )
```
Dreamsorcerer, line 213:
```suggestion
    return Cache(cache,  serializer=serializer,  plugins=plugins,  **cache_kwargs)
```

Dreamsorcerer:
As mentioned in that issue, doesn't look like the deprecation has actually happened yet. But, might as well update this.

PR #577, opened 2022-09-04
Add py.typed file in python package

titilambert:
## What do these changes do?

The file `py.typed` is present but it's not deployed/used when we use the latest commit.

This PR adds the missing "code" to add the `py.typed` file in the python package

The package will be see as typed package by mypy and alternatives

## Checklist

- [X] Unit tests for the changes exist
- [X] Documentation reflects the changes

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,1 @@
+global-include py.typed
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,1 +1,0 @@
-global-include py.typed
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,3 +1,7 @@
+[options.package_data]
+aiocache =
+    = py.typed
+
 [bdist_wheel]
 universal=1
 


titilambert:
I'm not sure about what I have to do for the `CHANGES` folder since it doesn't exists

Dreamsorcerer:
Thanks. I'd prefer to keep it consistent with our other libraries though. In aiohttp, it appears we just use include_package_data: https://github.com/aio-libs/aiohttp/blob/master/setup.cfg#L47
It also doesn't appear in MANIFEST.in, so not sure that's a requirement.

Dreamsorcerer:
It's not in aiohttp-jinja2 either, maybe the `graft` line does the job? https://github.com/aio-libs/aiohttp-jinja2/blob/master/MANIFEST.in#L8

titilambert:
I just tried with the `manifest.in` file, and when I install aiocache as dependency, it doesn't come with the `py.typed`. Which breaks mypy test
```
± mypy  hydroqc/
hydroqc/hydro_api/cache.py:6: error: Skipping analyzing "aiocache": module is installed, but missing library stubs or py.typed marker
hydroqc/hydro_api/cache.py:6: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
```

Or do you provide a lib stubs ?

Dreamsorcerer:
> I just tried with the `manifest.in` file, and when I install aiocache as dependency, it doesn't come with the `py.typed`. Which breaks mypy test

Did you try using a `graft` line like we have in other projects?
https://github.com/aio-libs/aiohttp-jinja2/blob/master/MANIFEST.in#L8

I'm guessing that is the one which would pull it in. I didn't mean that it currently works, just that I'd prefer the changes to match what we are doing in other repositories. I only setup mypy on this repository recently, so this is something I obviously missed.

Dreamsorcerer:
Done in #626, let me know if that's not working for you.

PR #584, opened 2022-12-27
Enable performance tests

Dreamsorcerer:


tests/performance/test_footprint.py
```diff
@@ -124,7 +124,7 @@ async def test_memcached_multigetsetdel(self, aiomcache_pool, memcached_cache):
             aiomcache_total_time += time.time() - start
 
         aiocache_total_time = 0
-        values = [b"a", b"b", b"c", b"d", b"e", b"f"]
+        values = ["a", "b", "c", "d", "e", "f"]
```
Dreamsorcerer, line 127: @argaen Any idea what is happening here?

The current tests fail because it tries to json decode these bytes objects:
https://github.com/aio-libs/aiocache/actions/runs/3789024806/jobs/6442383652

But, the `dumps` call in `multi_set()` appears to have existed before these tests were created, so I don't understand how these tests could have ever passed before.

Dreamsorcerer: I'll merge this as it has got the tests working, but let me know if you think this might indicate a regression or something.

PR #618, opened 2023-01-01
Add Python 3.11

Dreamsorcerer:


a68366:
Hello!
The tests are failing because of this change [in 3.11](https://docs.python.org/3/whatsnew/3.11.html#enum):

_Changed Enum.__format__() (the default for format(), str.format() and f-strings) of enums with mixed-in types (e.g. int, str) to also include the class name in the output, not just the member’s key. This matches the existing behavior of enum.Enum.__str__(), returning e.g. 'AnEnum.MEMBER' for an enum AnEnum(str, Enum) instead of just 'MEMBER'._

The [_build_key](https://github.com/aio-libs/aiocache/blob/84d92632efa5accc4ffb9e6912a5beaeba4df4ad/aiocache/base.py#L491) in 3.11 returns `Keys.KEY` instead of using [this enum ](https://github.com/aio-libs/aiocache/blob/master/tests/utils.py#L4)values.

If you just need to fix the tests, then something like this works: https://github.com/a68366/aiocache/pull/2 .

I'm not sure if just fixing the tests is enough though, since this change in python's behavior makes aiocache unreliable in an environment where 3.11 and lower versions coexist and enums are used for cache keys.

Dreamsorcerer:
@a68366 Thanks a lot for digging that out. The actual key values are not important, right? Only that they are consistent. So, what if we just do `str(key)` in `_build_key()`? That should produce the same result on all Python versions.

It'll invalidate existing caches when people upgrade to 0.12, but shouldn't cause any further issues after that.

a68366:
I'm not so familiar with aiocache codebase, but LGTM.

PR #619, opened 2023-01-02
Release 0.12

### LINKED ISSUE #536: New release?

Dreamsorcerer:
Fixes #536.


Dreamsorcerer:
@webknjaz Doesn't appear to be any deploy keys here. Can you set that up?

webknjaz:
@Dreamsorcerer yes, there are no secrets with the PyPI API keys in the repo. I've checked and realized that I don't have access to the PyPI project from my personal account or the bot service account. Only @asvetlov and `blck` seem to be listed there. Maybe, @argaen knows something?

webknjaz:
FWIW, I also don't have access to RTD. Only @argaen does.

argaen:
👋 I still have access to pypi and rtd, I could give access. @webknjaz @Dreamsorcerer should I give access to both of you?

About the keys.. if I have them they are in a laptop I dont have access right now so it's going to take some more time but my suggestion would maybe once you have access to pypi to create a new pair and store them properly

webknjaz:
@argaen yes, I think you could add me at least, as an Owner on PyPI and RTD. Maybe, both of us to fight the bus factor. Though, strictly speaking, it would be enough to give `aio-libs-bot` the Maintain privilege on PyPI and this would let me create a scoped token through that account. Thanks!

argaen:
@webknjaz I've sent invitations for both RTD and pypi

webknjaz:
@argaen thanks, I was able to add the bot there too.

webknjaz:
@Dreamsorcerer env secret provisioned. Also added a CNAME entry for RTD.

P.S. Looks like a few readme badges need updating.

CHANGELOG.md
```diff
@@ -1,19 +1,20 @@
 # CHANGELOG
 
 
-## 0.12.0 (2023-xx-xx)
+## 0.12.0 (2023-01-13)
 
 * Add ``async with`` support to ``BaseCache``.
```
webknjaz, line 6: FWIW it'd be nice to standardize on a consistent style natural to changelogs. Like describing what's changed since the last release in a non-imperative style... https://github.com/aio-libs/aiohttp/tree/master/CHANGES#alright-so-how-to-add-a-news-fragment

Dreamsorcerer:
@webknjaz Not sure if there's something else needed in the pipeline, but doesn't look like RTD got deployed (atleast the title still says 0.11.1).

webknjaz:
@Dreamsorcerer the last passed build was 7h 40min ago, the following ones all fail. See https://readthedocs.org/projects/aiocache/builds/.

webknjaz:
I've just enabled the PR builds feature so it'll get more visibility in the future. Plus, if you add the readme badge, it'll also show up red there.

Dreamsorcerer:
Right, thanks. Didn't realise it was building on each commit. That's easy enough to fix.

Issue #569, opened 2022-04-24
`build_key(key, namespace)` is missing from some modules/classes


pshafer-als:
There are some parts of `aiocache` that do not honor `namespace`s. This means that `BaseCache.build_key(key, namespace)` is never called to join the `namespace` with the `key`. The affected modules/classes are:
* `decorators.cached`
* `plugins.HitMissRatioPlugin`
* `backends.memory.SimpleMemoryCache`

This affects sharing an `alias`ed cache across multiple `@cached`-decorated callables. This currently works only when `namespace==None`.

A PR will be submitted shortly.

Dreamsorcerer:
Needs test/reproducer.

pshafer-als:
OK, I'll add these tests during the next few days.

pshafer-als:
### Here is a start...

I've added two new tests to "tests/acceptance/test_decorators.py::TestCached" in the [`test-namespace-key` branch of this fork](https://github.com/pshafer-als/aiocache/tree/test-namespace-key).
1. [`test_cached_without_namespace`](https://github.com/pshafer-als/aiocache/blob/d1290cb75f6e156e42fe7617d5aaee317b2a56a3/tests/acceptance/test_decorators.py#L59-L61) passes because the cache key is created with the default format when no namespace is provided.
2. [`test_cached_with_namespace`](https://github.com/pshafer-als/aiocache/blob/d1290cb75f6e156e42fe7617d5aaee317b2a56a3/tests/acceptance/test_decorators.py#L76-L78) fails because the default key for a decorated function/coroutine does not include the namespace.

### Run these tests:
```
pytest tests/acceptance/test_decorators.py::TestCached
```

pshafer-als:
The other changes included in https://github.com/aio-libs/aiocache/pull/570 are related to using cache keys that are not strings (notably `tuple` keys), so I will suggest those separately in a new issue.

@Dreamsorcerer if you agree that the tests above are worth resolving, I'll submit a new PR with just those changes to "decorators.py" and their corresponding tests.

pshafer-als:
> Tests look reasonable, go ahead and complete the PR.

Thanks.  Will do.

> I was under the impression that cache keys are expected to be strings always..

Yeah, in retrospect I'm seeing that now more than when I started using `aiocache` some months ago.  FWIW, Using non-string keys works quite well with the existing package when using a custom `key_builder` and the `key_filter` that I suggested in https://github.com/aio-libs/aiocache/pull/570.

[...]

PR #627, opened 2023-01-04
Decorator namespace key

pshafer-als:
### Keys with namespaces now work for decorators get/set
* Include build_key(key, namespace) in decorators.cached:
    * get_from_cache()
    * set_in_cache()

* Keys with namespaces now work for the @cached() decorator

## Related issue number
* https://github.com/aio-libs/aiocache/issues/569

## Checklist

- [x] Add a new news fragment into the `CHANGES` folder

### DIFF
--- a/aiocache/decorators.py
+++ b/aiocache/decorators.py
@@ -177,6 +177,11 @@ class cached_stampede(cached):
         key_from_attr param. If key and key_from_attr are not passed, it will use module_name
         + function_name + args + kwargs
     :param key_from_attr: str arg or kwarg name from the function to use as a key.
+    :param namespace: string to use as default prefix for the key used in all operations of
+        the backend. Default is None
+    :param key_builder: Callable that allows to build the function dynamically. It receives
+        the function plus same args and kwargs passed to the function.
+        This behavior is necessarily different than ``BaseCache.build_key()``
     :param cache: cache class to use when calling the ``set``/``get`` operations.
         Default is :class:`aiocache.SimpleMemoryCache`.
     :param serializer: serializer instance to use when calling the ``dumps``/``loads``.
@@ -201,7 +206,8 @@ class cached_stampede(cached):
         if value is not None:
             return value
 
-        async with RedLock(self.cache, key, self.lease):
+        async with RedLock(
+                self.cache, key, self.lease, namespace=self._namespace):
             value = await self.get_from_cache(key)
             if value is not None:
                 return value
--- a/aiocache/decorators.py
+++ b/aiocache/decorators.py
@@ -309,6 +309,9 @@ class multi_cached:
     def __call__(self, f):
         if self.alias:
             self.cache = caches.get(self.alias)
+            for arg in ("serializer", "namespace", "plugins"):
+                if getattr(self, f'_{arg}', None) is not None:
+                    logger.warning(f"Using cache alias; ignoring '{arg}' argument.")
         else:
             self.cache = _get_cache(
                 cache=self._cache,
--- a/aiocache/decorators.py
+++ b/aiocache/decorators.py
@@ -86,6 +86,9 @@ class cached:
     def __call__(self, f):
         if self.alias:
             self.cache = caches.get(self.alias)
+            for arg in ("serializer", "namespace", "plugins"):
+                if getattr(self, f'_{arg}', None) is not None:
+                    logger.warning(f"Using cache alias; ignoring '{arg}' argument.")
         else:
             self.cache = _get_cache(
                 cache=self._cache,
--- a/aiocache/decorators.py
+++ b/aiocache/decorators.py
@@ -135,16 +135,16 @@ class cached:
         )
 
     async def get_from_cache(self, key):
+        namespace = self._kwargs.get("namespace", None)
         try:
-            namespace = self._kwargs.get("namespace", None)
             return await self.cache.get(key, namespace=namespace)
         except Exception:
             logger.exception("Couldn't retrieve %s, unexpected error", key)
         return None
 
     async def set_in_cache(self, key, value):
+        namespace = self._kwargs.get("namespace", None)
         try:
-            namespace = self._kwargs.get("namespace", None)
             await self.cache.set(key, value, namespace=namespace, ttl=self.ttl)
         except Exception:
             logger.exception(
@@ -209,9 +209,9 @@ class cached_stampede(cached):
 
 def _get_cache(cache=Cache.MEMORY, serializer=None, plugins=None, **cache_kwargs):
     return Cache(
-        cache, 
-        serializer=serializer, 
-        plugins=plugins, 
+        cache,
+        serializer=serializer,
+        plugins=plugins,
         **cache_kwargs,
     )
 
--- a/aiocache/decorators.py
+++ b/aiocache/decorators.py
@@ -38,8 +38,11 @@ class cached:
     :param key: str value to set as key for the function return. Takes precedence over
         key_builder param. If key and key_builder are not passed, it will use module_name
         + function_name + args + kwargs
+    :param namespace: string to use as default prefix for the key used in all operations of
+        the backend. Default is None
     :param key_builder: Callable that allows to build the function dynamically. It receives
         the function plus same args and kwargs passed to the function.
+        This behavior is necessarily different than ``BaseCache.build_key()``
     :param cache: cache class to use when calling the ``set``/``get`` operations.
         Default is :class:`aiocache.SimpleMemoryCache`.
     :param serializer: serializer instance to use when calling the ``dumps``/``loads``.
@@ -58,6 +61,7 @@ class cached:
         self,
         ttl=SENTINEL,
         key=None,
+        namespace=None,
         key_builder=None,
         cache=Cache.MEMORY,
         serializer=None,
@@ -75,6 +79,7 @@ class cached:
 
         self._cache = cache
         self._serializer = serializer
+        self._namespace = namespace
         self._plugins = plugins
         self._kwargs = kwargs
 
@@ -85,6 +90,7 @@ class cached:
             self.cache = _get_cache(
                 cache=self._cache,
                 serializer=self._serializer,
+                namespace=self._namespace,
                 plugins=self._plugins,
                 **self._kwargs,
             )
@@ -135,7 +141,7 @@ class cached:
         )
 
     async def get_from_cache(self, key):
-        namespace = self._kwargs.get("namespace", None)
+        namespace = self.cache.namespace
         try:
             return await self.cache.get(key, namespace=namespace)
         except Exception:
@@ -143,7 +149,7 @@ class cached:
         return None
 
     async def set_in_cache(self, key, value):
-        namespace = self._kwargs.get("namespace", None)
+        namespace = self.cache.namespace
         try:
             await self.cache.set(key, value, namespace=namespace, ttl=self.ttl)
         except Exception:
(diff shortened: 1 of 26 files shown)


aiocache/decorators.py
```diff
@@ -134,16 +134,18 @@ def _key_from_args(self, func, args, kwargs):
             + str(ordered_kwargs)
         )
 
-    async def get_from_cache(self, key: str):
+    async def get_from_cache(self, key):
         try:
-            return await self.cache.get(key)
+            namespace = self._kwargs.get("namespace", None)
```
Dreamsorcerer, line 139: Presumably, this shouldn't cause an exception, so move it before the `try`.

aiocache/decorators.py
```diff
@@ -134,16 +134,18 @@ def _key_from_args(self, func, args, kwargs):
...
+            return await self.cache.get(key, namespace=namespace)
         except Exception:
             logger.exception("Couldn't retrieve %s, unexpected error", key)
         return None
 
     async def set_in_cache(self, key, value):
         try:
-            await self.cache.set(key, value, ttl=self.ttl)
+            namespace = self._kwargs.get("namespace", None)
```
Dreamsorcerer, line 147: Same

aiocache/decorators.py
```diff
@@ -205,7 +207,12 @@ async def decorator(self, f, *args, **kwargs):
 
 
 def _get_cache(cache=Cache.MEMORY, serializer=None, plugins=None, **cache_kwargs):
-    return Cache(cache, serializer=serializer, plugins=plugins, **cache_kwargs)
+    return Cache(
```
Dreamsorcerer, line 211: Looks unrelated (maybe I need to tweak a Black config or something...).

tests/acceptance/test_decorators.py
```diff
@@ -48,6 +48,35 @@ async def fn(self, a, b=2):
...
+        """Default cache key is created when no namespace is provided"""
+        @cached(namespace=None)
+        async def fn():
+            return "1"
+        decorated_fn = cached(fn, namespace=None)
+        
+        await fn()
+        assert await cache.exists(
+            decorated_fn.get_cache_key(fn, args=(), kwargs={})
+            ) is True
```
Dreamsorcerer, line 61:
```suggestion
        key = decorated_fn.get_cache_key(fn, args=(), kwargs={})
        assert await cache.exists(key) is True
```

tests/acceptance/test_decorators.py
```diff
@@ -48,6 +48,35 @@ async def fn(self, a, b=2):
...
+            return "1"
+        decorated_ns_fn = cached(ns_fn, namespace=key_prefix)
+        
+        await ns_fn()
+        assert await cache.exists(
+            decorated_ns_fn.get_cache_key(ns_fn, args=(), kwargs={})
+            ) is True
+        assert decorated_ns_fn.get_cache_key(
+            ns_fn, args=(), kwargs={},
+            ).startswith(key_prefix)
```
Dreamsorcerer, line 78:
```suggestion
        key = decorated_ns_fn.get_cache_key(ns_fn, args=(), kwargs={})
        assert await cache.exists(key) is True
        assert key.startswith(key_prefix)
```

aiocache/decorators.py
pshafer-als: I've deleted space characters at ends of lines here.

Dreamsorcerer:
Tests are failing.

pshafer-als:
> Tests are failing.

Noted.  Changes are in progress now...I'm planning to push the corrections within the next hour.

Dreamsorcerer:
Nope, still failing. You're not testing in 3.7 are you? Because only half the tests run on 3.7.

pshafer-als:
I've now updated all decorators to use the namespace if provided.

To make this work, I updated `TestCachedStampede.test_cached_stampede` to use the namespace in the call signature of the mock `cache.get()` and `cache.set()`.

This is the intended behavior after the changes I made in this branch to the `decorators.cached`. But I don't know whether there was a good reason to explicitly exclude the namespace in this test previously.

I made similar changes in tests/ut/ so that `mock_cache` uses `namespace="test"`.

aiocache/decorators.py
```diff
@@ -75,6 +79,7 @@ def __init__(
 
         self._cache = cache
         self._serializer = serializer
+        self._namespace = namespace
```
Dreamsorcerer, line 82: In some places, you've used `self.cache.namespace`, others `self._namespace`, do we need this extra attribute?

aiocache/decorators.py
```diff
@@ -134,18 +140,21 @@ def _key_from_args(self, func, args, kwargs):
             + str(ordered_kwargs)
         )
 
-    async def get_from_cache(self, key: str):
+    async def get_from_cache(self, key):
+        namespace = self.cache.namespace
         try:
-            return await self.cache.get(key)
+            return await self.cache.get(key, namespace=namespace)
```
Dreamsorcerer, line 146: Isn't the cache already using it's own namespace?

Dreamsorcerer: It looks to me like the build_key method is expected to use `self.namespace` if `None`:
https://github.com/pshafer-als/aiocache/blob/decorator-namespace-key/aiocache/base.py#L499

aiocache/decorators.py
```diff
@@ -38,8 +38,11 @@ class cached:
     :param key: str value to set as key for the function return. Takes precedence over
         key_builder param. If key and key_builder are not passed, it will use module_name
         + function_name + args + kwargs
+    :param namespace: string to use as default prefix for the key used in all operations of
+        the backend. Default is None
     :param key_builder: Callable that allows to build the function dynamically. It receives
         the function plus same args and kwargs passed to the function.
+        This behavior is necessarily different than ``BaseCache.build_key()``
```
Dreamsorcerer, line 45: I'm looking at the code, and not really seeing the point of this `key_builder`. Maybe we should just remove it and use the cache methods (later)...

aiocache/decorators.py
Dreamsorcerer: I don't see a single example of how to use a custom build_key, so maybe it'd make more sense if all the calls to `self.build_key()` did `namespace=namespace or self.namespace`. Then it's more likely to work as expected.

I think this is potentially backwards-incompatible, so I'd prefer if we did this change in a later PR. Can we get this working without passing the namespace in? Even if a couple of tests need to have xfail on them until that change is made.

aiocache/decorators.py
pshafer-als: Probably not.  Comments below suggest that it might not be needed after __init__() completes.

aiocache/decorators.py
pshafer-als: Yes, that was on my mind also.  Some of this is remnant of a shift in strategy to better use the underlying case, rather than implement namespace functionality directly in the decorator.

I'll clean this up and see how the test perform.

aiocache/decorators.py
pshafer-als: This `key_builder` in the decorators is necessary to make a cache key from the call signature of the function/coroutine that is being decorated.  It's the only place where the "raw" key gets generated. It accepts the callable object and the `args` + `kwargs` from that callable.

Whereas the `key_builder` supplied to `BaseCache` (which becomes `BaseCache.build_key`) accepts only two parameters: `key` and `namespace`, and it concatenates them by default. The "raw" key must already exist.

Perhaps these two functionalities of theses parameters should have separate names, rather than both being designated as `key_builder`?

Dreamsorcerer: OK, let's leave it alone for now then. Can probably do with some better documentation or something later though.

Dreamsorcerer:
Looks like we have what we need to make a new release now. Think you can finish this off in the next couple of days?

pshafer-als:
> Looks like we have what we need to make a new release now. Think you can finish this off in the next couple of days?

Yes, I can make that work.

pshafer-als:
In reference to [this comment](https://github.com/aio-libs/aiocache/issues/569#issuecomment-1376107644) in Issue #569, here is a summary of the responses included in this PR.

1.  N/A
2. ...
    a. See examples/alt_key_builder.py and updates to examples/cached_decorator.py
    b. `BaseCache.build_key` uses `namespace` argument if provided, otherwise it uses `self.namespace`.
    c. Cache locks use `client.build_key` rather than `client._build_key`.
3. ...
    a. _Deferred_. This might get submitted as a separate issue / PR at a later time.
    b. `namespace` is an explicit parameter to decorator caches.
4. _Deferred_. This might get submitted as a separate issue / PR at a later time.
5. Warn when decorator parameters are unused because alias takes precedence.
6. _Deferred_. This might get submitted as a separate issue / PR at a later time.
7. N/A

pshafer-als:
@Dreamsorcerer  I think this is ready to go, but let me know if you see additional changes needed.

examples/alt_key_builder.py
```diff
@@ -0,0 +1,293 @@
+"""alt_key_builder.py
```
Dreamsorcerer, line 1: This file feels a little too verbose/dense for an example, but I'll merge it for now and look at it again in a later release.

tests/acceptance/test_lock.py
```diff
@@ -97,6 +115,26 @@ class TestRedisRedLock:
     def lock(self, redis_cache):
         return RedLock(redis_cache, Keys.KEY, 20)
 
+    @pytest.fixture
+    def set_key_builder(self, redis_cache):
```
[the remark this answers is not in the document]
pshafer-als: DONE

examples/cached_decorator.py
```diff
@@ -16,12 +17,56 @@ async def cached_call():
...
+        'serializer': {
+            'class': StringSerializer,
+        },
+        'namespace': "demo",
+        'key_builder': build_key,
+    },
+)
+
+
+async def test_cached_alias_build_key():
```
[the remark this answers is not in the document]
pshafer-als: DONE

pshafer-als (commented):
All comments appear to be resolved

pshafer-als:
@Dreamsorcerer Please have another look, and let me know if all is good.

Dreamsorcerer:
Now, if you're happy to do some more, it would be great if you can create a PR which reverts all the extra namespace conversions in here and does it something like this:
```
def __init__(self, key_builder):
    self._build_key = key_builder or lambda k, ns: k if ns is None else "{}{}".format(ns, k)

def build_key(self, key, namespace=None):
    return self._build_key(key, namespace or self.namespace)
```

Then we can do it like this for v1 and introduce some backwards incompatible changes.

Should also be able to fold in the `_ensure_key()` logic and remove that function. Plus, MemcachedBackend can do the `.encode()` in `build_key()` as well.

In other words, the key_builder should not need to know about any of these edge cases, it should just receive a `str` and return a `str`.

pshafer-als:
Nice!  I like those ideas.  

I should have some time over the next couple weeks to give this a try.

Dreamsorcerer:
@pshafer-als Any chance you'll get round to that soon?

pshafer-als:
That's a fair question.  Realistically, I think it will be ~2 weeks before I can secure sufficient time to work on this in earnest.  I think it will be a few hours of work, but probably spread over a 1-2 week period.

If you need to move on, I understand. Otherwise, I'm still interested.

Dreamsorcerer:
That's good, just making sure the task doesn't get lost.

pshafer-als:
Started a [branch for this work](https://github.com/pshafer-als/aiocache/blob/simplify-cache-namespace/aiocache/base.py), and a [new thread for discussion](https://github.com/aio-libs/aiocache/issues/669).

PR #634, opened 2023-01-10
Use f-string to eliminate cast to `str` for 3.11 compatibility

### LINKED ISSUE #633: Ensure that underlying value of enum key is used by BaseCache._build_key() across Python versions
Key builders and tests currently cast key values (notably `Enum` values) to `str` to maintain compatibility with Python 3.11 and older versions.  ~~This requirement can be eliminated by using f-strings rather than `str.__format__`.~~

EDIT: Whoops, that erroneous statement was based on testing in the wrong python environment.  Instead of using `str()`, perhaps we should use a simple guard that check for an enum key.  Something like:

```
def ensure_key(key):
    if isinstance(key, Enum):
        return key.value
    else:
        return key
```

pshafer-als:
## Use f-string to ensure that keys are strings

This eliminates the need to cast all keys (especially Enum values) to `str` in conjunction with `str.format()`.

Keys are named as intended, using the key variable's value rather than using the name of the key variable.

## Related issue number

Resolves #633.

## Checklist

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,2 @@
+[run]
+concurrency = multiprocessing
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,7 +50,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu]
-        pyver: ['3.7', '3.8', '3.9', '3.10']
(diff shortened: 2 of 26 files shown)


pshafer-als:
Withdrawing this PR.  Will submit a more suitable PR in response to #633.

PR #635, opened 2023-01-10
Use `base.ensure_key()` in `_build_key()` to ensure consistency of enum keys between different Python versions

### LINKED ISSUE #633: Ensure that underlying value of enum key is used by BaseCache._build_key() across Python versions

pshafer-als:
## New ensure_key() works with enum keys

`BaseCache._build_key()` uses `ensure_key(key)` to replace enum keys with the Enum's value (rather than the Enum object, as now happens in Python 3.11).

Enum keys (or string keys) can be used consistently across different Python versions.

## Related issue number

Resolves #633.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@
 * Add ``async with`` support to ``BaseCache``.
 * Remove deprecated ``loop`` parameters.
 * Remove deprecated ``cache`` parameter from ``create()``.
-* Use ``str()`` in ``_build_key()`` to ensure consistency of enum keys between different Python versions (if using enum keys, upgrading to 0.12 may invalidate existing caches due to key values changing).
+* Use ``base._ensure_key()`` in ``_build_key()`` to ensure consistency of enum
+keys between different Python versions. [#633](https://github.com/aio-libs/aiocache/issues/633) -- Padraic Shafer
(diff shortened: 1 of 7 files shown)


pshafer-als:
> Ah, I think you branched off your other PR. Unfortunately, we always squash commit, so that's messed up the diff/conflicts.
> 
> If you can rename it to `_ensure_key()`, as I'd like to replace it in the next release, and clean up the merge, that'd be great. We can push a new release once this is merged.

No problem.  Will do that now.

pshafer-als:
> If you can rename it to `_ensure_key()`, as I'd like to replace it in the next release, and clean up the merge, that'd be great. We can push a new release once this is merged.

Done.  

Thank you for nudging these tests along and for all your help.  It's been a pleasure working with you on these PRs.

PR #652, opened 2023-01-20
Introduce skip_cache_func

### LINKED ISSUE #649: Block writing cache in `cached` decorator if custom condition is not satisfied
I m using the  `aiocache.cached` decorator with wraps the certain processing function. The setup is as following:

```python
import aiocache


aiocache.caches.add(
    "cache-alias",
    {
        "cache": "aiocache.RedisCache",
        {
            <cache-config-data>
        }
    },
)


@aiocache.cached(
    alias="cache-alias",
    key_builder=some_key_builder,
    ttl=ttl,
)
async def processing(...) -> int:
    ...
```

The problem is -- can i pass any argument to cached decorator in order to avoid cache writing for `processing(...)` return?
It seems like the only way to perform it is to throw exception inside the `processing(...)` function if the result is not appropriate one. I m looking for more generic way to do the same stuff.

For example, pass some `Callable` object to `cached.__call__()` returning the bool var meaning caching the results. Something similar to:

```python
def checker(value: Any) -> bool:
    ...

@aiocache.cached(
    alias="cache-alias",
    key_builder=some_key_builder,
    value_checker=checker,
    ttl=ttl,
)
async def processing(...) -> int:
    ...
```

VitalyPetrov:
- add an optional `value_checker` param to cache decorator classes `cached`, `cached_stampede` and `multi_cached`
- add IDE-specific folders to `gitignore`
- add black formatter to `requirements-dev.txt

Did not test in real production code

## Related issue number

Fixes #649

## Checklist

- [+] I think the code is well written
- [-] Unit tests for the changes exist
- [+] Documentation reflects the changes
- [-] Add a new news fragment into the `CHANGES` folder

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -86,6 +86,10 @@ ENV/
 # Spyder project settings
 .spyderproject
 
+# IDE project files
+.idea/
+.vscode/
+
 # Rope project settings
 .ropeproject
 
--- a/.gitignore
+++ b/.gitignore
@@ -86,10 +86,6 @@ ENV/
 # Spyder project settings
 .spyderproject
 
-# IDE project files
-.idea/
-.vscode/
-
 # Rope project settings
 .ropeproject
 
--- a/aiocache/decorators.py
+++ b/aiocache/decorators.py
@@ -39,6 +39,10 @@ class cached:
     :param key_builder: Callable that allows to build the function dynamically. It receives
         the function plus same args and kwargs passed to the function.
         This behavior is necessarily different than ``BaseCache.build_key()``
+    :param value_checker: Callable that allows to control whether the return
+        value to be cached or not. The callable receives the return value of the
+        wrapped function and should return the boolean indicating whether
+        to cache the result or not.
     :param cache: cache class to use when calling the ``set``/``get`` operations.
         Default is :class:`aiocache.SimpleMemoryCache`.
     :param serializer: serializer instance to use when calling the ``dumps``/``loads``.
@@ -58,6 +62,7 @@ class cached:
         ttl=SENTINEL,
         namespace=None,
         key_builder=None,
+        value_checker=None,
         cache=Cache.MEMORY,
         serializer=None,
         plugins=None,
@@ -67,6 +72,7 @@ class cached:
     ):
         self.ttl = ttl
         self.key_builder = key_builder
+        self.value_checker = value_checker
         self.noself = noself
         self.alias = alias
         self.cache = None
@@ -110,6 +116,13 @@ class cached:
                 return value
 
         result = await f(*args, **kwargs)
+
+        if self.value_checker:
+            if not isinstance(self.value_checker(result), bool):
+                raise TypeError('value_checker callable should return bool')
+
+            if not self.value_checker(result):
+                return result
 
         if cache_write:
             if aiocache_wait_for_write:
@@ -171,6 +184,10 @@ class cached_stampede(cached):
     :param key_builder: Callable that allows to build the function dynamically. It receives
         the function plus same args and kwargs passed to the function.
         This behavior is necessarily different than ``BaseCache.build_key()``
+    :param value_checker: Callable that allows to control whether the return
+        value to be cached or not. The callable receives the return value of the
+        wrapped function and should return the boolean indicating whether
+        to cache the result or not.
     :param cache: cache class to use when calling the ``set``/``get`` operations.
         Default is :class:`aiocache.SimpleMemoryCache`.
     :param serializer: serializer instance to use when calling the ``dumps``/``loads``.
@@ -201,6 +218,13 @@ class cached_stampede(cached):
                 return value
 
             result = await f(*args, **kwargs)
+
+            if self.value_checker:
+                if not isinstance(self.value_checker(result), bool):
+                    raise TypeError('value_checker callable should return bool')
+
+                if not self.value_checker(result):
+                    return result
 
             await self.set_in_cache(key, result)
 
--- a/aiocache/decorators.py
+++ b/aiocache/decorators.py
@@ -375,7 +375,7 @@ class multi_cached:
         result = await f(*new_args, **kwargs)
         result.update(partial)
 
-        result2cache = {k: v for k, v in result.items() if self.skip_cache_func(k, v)}
+        result2cache = {k: v for k, v in result.items() if not self.skip_cache_func(k, v)}
 
         if not result2cache:
             return result
--- a/aiocache/decorators.py
+++ b/aiocache/decorators.py
@@ -375,17 +375,17 @@ class multi_cached:
         result = await f(*new_args, **kwargs)
         result.update(partial)
 
-        result2cache = {k: v for k, v in result.items() if not self.skip_cache_func(k, v)}
+        to_cache = {k: v for k, v in result.items() if not self.skip_cache_func(k, v)}
 
-        if not result2cache:
+        if not to_cache:
             return result
 
         if cache_write:
             if aiocache_wait_for_write:
-                await self.set_in_cache(result2cache, f, args, kwargs)
+                await self.set_in_cache(to_cache, f, args, kwargs)
             else:
                 # TODO: Use aiojobs to avoid warnings.
-                asyncio.create_task(self.set_in_cache(result2cache, f, args, kwargs))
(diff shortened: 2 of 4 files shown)


VitalyPetrov:
Unit tests for that functionality is not done yet.
As for notes in `CHANGES` the PR seems to be too small to be released separately IMO. But the choice Is yours (maintainers).

aiocache/decorators.py
```diff
@@ -58,6 +62,7 @@ def __init__(
         ttl=SENTINEL,
         namespace=None,
         key_builder=None,
+        value_checker=None,
```
Dreamsorcerer, line 65:
```suggestion
        skip_cache_func=lambda r: False,
```

How about we just put a default function and not need to test for `None`?

aiocache/decorators.py
```diff
@@ -111,6 +117,13 @@ async def decorator(
 
         result = await f(*args, **kwargs)
 
+        if self.value_checker:
+            if not isinstance(self.value_checker(result), bool):
+                raise TypeError('value_checker callable should return bool')
```
Dreamsorcerer, line 122: I'd be inclined to leave this to type checkers (e.g. mypy) and not care about runtime checks here.

requirements-dev.txt
```diff
@@ -1,5 +1,6 @@
 -r requirements.txt
 
+black==22.12.0
```
Dreamsorcerer, line 3: Not relevant to the PR.

```suggestion
```

aiocache/decorators.py
```diff
@@ -268,6 +292,11 @@ class multi_cached:
         ``keys_from_attr``, the decorated callable, and the positional and keyword arguments
         that were passed to the decorated callable. This behavior is necessarily different than
         ``BaseCache.build_key()`` and the call signature differs from ``cached.key_builder``.
+    :param value_checker: dict of callables that allows to control which keys should not
+        be cached based on corresponding value check. The dict keys are set of keys to
+        be checked while the dict value is callable which takes corresponding result value
+        (result[key])  and returns boolean indicating whether cache that (key, value)
+        pair or not.
```
Dreamsorcerer, line 299: Hmm, I'm not sure about this one, it seems to be less flexible than the others.
This requires knowing which keys should (or could) be cached. e.g. You couldn't skip keys that ended with `_nocache`, for example, unless you knew all possible keys beforehand.

I can think of 2 flexible options:
```
skip_keys = skip_cache_func(result.copy())
results_to_cache = {k: v for k, v in result.items() if k not in skip_keys}

# OR

results_to_cache = dict(it for it in result.items() if not skip_cache_func(*it))
```

Dreamsorcerer (commented):
Thanks.

A couple of suggestions there, and we'll also need some tests for the new behaviour (maybe in tests/acceptance/tests_decorator.py so it gets tested against the different backends).

Dreamsorcerer:
Ignore the news fragment. I'm not sure where that checklist comes from, trying to see if I can edit it...

aiocache/decorators.py
VitalyPetrov: Its an antipattern
https://www.flake8rules.com/rules/E731.html

Dreamsorcerer: The function returns True, it's never going to appear in a traceback..

.gitignore
```diff
@@ -86,6 +86,10 @@ ENV/
 # Spyder project settings
 .spyderproject
 
+# IDE project files
+.idea/
+.vscode/
+
```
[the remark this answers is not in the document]
VitalyPetrov: Yes, that fine, but allows from other dev to make the same settings on his machine. Local `.gitignore` is more flexible  here as you can be sure that there will be no extra files in your repo.

aiocache/decorators.py
Dreamsorcerer: Also, the antipattern is assigning to a variable. lambdas exist to be passed directly into a function.
`foo(lambda: ...)` is allowed
`f = lambda: ...` is not

.gitignore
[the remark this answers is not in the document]
Dreamsorcerer: Expecting every project in the world to add ignore patterns for every IDE and tool in the world is not more flexible (but, seems to be what most people expect with Git). Global ignore all files for _your_ editors and you'll never commit them into another repo again.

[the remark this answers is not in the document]
Dreamsorcerer: What I mean is that gitignore in a repo should be related to that repo. `.vscode` has nothing to do with the project, it's related to someone's computer, and therefore should be ignored on their computer (if VS code were smarter, it might add this to a global ignore when it was installing itself).

The gitignore files we have at work are typically about 2 lines long. They just contain something like `/dist` which is created by the build tool in our repo. More importantly, these changes have nothing to do with 'Feature/value checker', it can just be a little frustrating when people try to sneak in stylistic changes etc. to a project in a random PR.

aiocache/decorators.py
Dreamsorcerer: So, my understanding is that when you do `f = lambda: ...`, you are essentially trying to give a name to a lambda (`f` in this example). But, when printing it out etc. you won't see that name. That is an antipattern.

In this case, we are not trying to assign a name to the lambda, it is the default value which would not have a name associated with it.

aiocache/decorators.py
VitalyPetrov: This approach also require you to know all the possible keys.
The original idea was to perform value-based filtering for resulting dict. However I am also not sure this functionality is needed

.gitignore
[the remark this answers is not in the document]
VitalyPetrov: Removed from `.gitignore`

aiocache/decorators.py
```diff
@@ -39,6 +39,10 @@ class cached:
     :param key_builder: Callable that allows to build the function dynamically. It receives
         the function plus same args and kwargs passed to the function.
         This behavior is necessarily different than ``BaseCache.build_key()``
+    :param value_checker: Callable that allows to control whether the return
+        value to be cached or not. The callable receives the return value of the
+        wrapped function and should return the boolean indicating whether
+        to cache the result or not.
```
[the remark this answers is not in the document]
VitalyPetrov: Agree, looks more clear.
Replaced

aiocache/decorators.py
```diff
@@ -111,6 +117,13 @@ async def decorator(
 
         result = await f(*args, **kwargs)
 
+        if self.value_checker:
+            if not isinstance(self.value_checker(result), bool):
+                raise TypeError('value_checker callable should return bool')
+
+            if not self.value_checker(result):
```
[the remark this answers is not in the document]
VitalyPetrov: Done

aiocache/decorators.py
VitalyPetrov: Done

aiocache/decorators.py
VitalyPetrov: The original idea was to provide an interface to perform value-based filtering (similar to your provided examples). So in both cases you need to know a range of all possible keys. However I m also not sure if this functionality is needed

VitalyPetrov:
Updated the code, please check

aiocache/decorators.py
Dreamsorcerer: I don't see why you need the possible keys? In the above examples my cache function can look like:
```
def f(results):
    return (k for k in results.keys() if k.endswith("_nocache"))
```
or
```
def f(k, v):
    return k.endswith("_nocache")
```

Whereas the original does not allow filtering keys ending with `_nocache` unless you can name every single possible key beforehand.

Dreamsorcerer: I'm feeling like the 2nd option is easier and clearer to understand, so I'm leaning towards that solution.

Dreamsorcerer: Obviously, you can just as easily check values without caring about keys:
```
def f(k, v):
    return v is None
```

VitalyPetrov: It seems I ve finally got your idea. Agree -- your variant is more generic and easy-to-use.

Dreamsorcerer:
OK, looking good. If we can just get a couple of tests in (maybe in tests/acceptance/tests_decorator.py?), then we're good to merge.

Dreamsorcerer:
Added cached_stampede to the tests to get full coverage, all good now.

Thanks a lot for that.

Issue #669, opened 2023-02-18
Simplify cache namespace and key encoding logic for v1


padraic-shafer:
> _Originally posted by @Dreamsorcerer in https://github.com/aio-libs/aiocache/issues/627#issuecomment-1377682413_
> ...create a PR which reverts all the extra namespace conversions in here and does it something like this:
> ```
> def __init__(self, key_builder):
>     self._build_key = key_builder or lambda k, ns: "{}{}".format(ns or "", k)
> 
> def build_key(self, key, namespace=None):
>     return self._build_key(key, namespace or self.namespace)
> ```

[Moving the namespace logic into a single function](https://github.com/pshafer-als/aiocache/blob/simplify-cache-namespace/aiocache/base.py) (`BaseCache.buildkey()` rather than `BaseCache.add()`, `get()` ,etc.) makes a lot of sense.  

This change causes tests/ut/test_base.py::TestBaseCache::test_alt_build_key_override_namespace[None-key] to fail: (`AssertionError: assert 'test:key' == 'key'`). This makes sense because the namespace gets overridden by `self.namespace` before the custom `key_builder` gets called. The strange part is that I think this test should have already failed before I made these changes.

I'll dig into that and the knock-on impacts to backends and decorators in a few days.

pshafer-als:
> This change causes tests/ut/test_base.py::TestBaseCache::test_alt_build_key_override_namespace[None-key] to fail: (`AssertionError: assert 'test:key' == 'key'`). This makes sense because the namespace gets overridden by `self.namespace` before the custom `key_builder` gets called. The strange part is that I think this test should have already failed before I made these changes.

Oh, right.  That's because this test is making a bare call to `build_key()`, rather than going through one of the other methods (`add`/`set`/`get`/...).  In the main branch, a custom `key_builder` has the ability to (optionally) override the namespace logic (including default value) of the cache.  The changes in this new branch put that control back to the cache, so that the only way to override the cache's default namespace is to provide an _explicit_ namespace when calling `BaseCache.build_key()`.

My hunch is that this will break the ability to override the default namespace in a cache decorator that is backed by an `alias` cache, and have a similar effect on other classes that don't explicitly accept namespace arguments (like `RedLock`).  Currently I have a project that relies on being able to override the namespace in this scenario.

I'll need to look into how best to resolve this.

pshafer-als:
[...]

Issue #625, opened 2023-01-02
Wrap functions in class with decorator


Dreamsorcerer:
The decorator should be simplified to a simple function, which should then return an object which wraps the passed function. Something along the lines of:
```
class Wrapper:
    def __init__(self, func):
        self.func = func

    def __call__(self, *args, **kwargs):
        # Get from cache...
        # Or call self.func(*args, **kwargs)

def cached(func):
    return Wrapper(self.func)
```

This simplifies the logic, and makes it easy to add methods on to our function (related: #538). For example, to call the function while forcing the cache to update could look something like:
```
@cached
def foo(...):
    ...

# Force update
foo.refresh(...)
```

slingshotvfx:
+1 for this feature, it would be great (especially for unit testing) to be able to call a function without the cache interfering, just like [async-lru](https://github.com/aio-libs/async-lru)'s `cache_invalidate()` method

mavwolverine:
I think the `cache_invalidate` should work on two levels:
1. Individual function call arguments - `get_user(user_id)` if the user object was updated in the DB.
2. Namespace level - if there was some bulk change to a particular type of data.

PR #670, opened 2023-02-19
Simplify cache namespace and key encoding logic for v1

pshafer-als:
## What do these changes do?

* All logic to select the namespace for a cache key, and encode the resulting namespaced key are now encapsulated in the `build_key()` member of `BaseCache` and its backend subclasses.
* Simplifies the burden of creating a custom ``key_builder`` for a cache.
* Consistent use of namespaced keys across backends and helpers (e.g., locks) while still allowing customization of the formatting and encoding of the namespaced key.

* When creating a cache object with a custom ``key_builder`` argument...
    * You no longer need to consider the backend encoding requirements for key identifiers; these are handled by the backend cache class.
    * The ``namespace`` argument passed to your custom ``key_builder`` has already been resolved by the cache: It will be the namespace that was explicitly requested by a client (if requested); otherwise it will be the default namespace that was used to instantiate the cache.
* If your code was calling the private `_build_key()` member of a cache, replace those calls with `build_key()`.

## Related issue number

Resolves issue #669.

## Checklist

### DIFF
--- a/aiocache/backends/memcached.py
+++ b/aiocache/backends/memcached.py
@@ -147,8 +147,8 @@ class MemcachedCache(MemcachedBackend):
     def parse_uri_path(cls, path):
         return {}
 
-    def _build_key(self, key, namespace=None):
-        ns_key = super()._build_key(key, namespace=namespace).replace(" ", "_")
+    def build_key(self, key, namespace=None):
+        ns_key = super().build_key(key, namespace=namespace).replace(" ", "_")
         return str.encode(ns_key)
 
     def __repr__(self):  # pragma: no cover
--- a/aiocache/backends/memcached.py
+++ b/aiocache/backends/memcached.py
@@ -1,5 +1,5 @@
 import asyncio
-from typing import cast, Optional, Union
+from typing import Optional
 
 import aiomcache
 
@@ -7,8 +7,7 @@ from aiocache.base import BaseCache
 from aiocache.serializers import JsonSerializer
 
 
-# class MemcachedBackend(BaseCache[str]):
-class MemcachedBackend(BaseCache):
+class MemcachedBackend(BaseCache[bytes]):
     def __init__(self, endpoint="127.0.0.1", port=11211, pool_size=2, **kwargs):
         super().__init__(**kwargs)
         self.endpoint = endpoint
@@ -121,7 +120,7 @@ class MemcachedBackend(BaseCache):
         await self.client.close()
 
 
-# class MemcachedCache(Generic[CacheKeyType], MemcachedBackend):
+# class MemcachedCache(MemcachedBackend):
 class MemcachedCache(MemcachedBackend):
     """
     Memcached cache implementation with the following components as defaults:
@@ -142,9 +141,7 @@ class MemcachedCache(MemcachedBackend):
     """
 
     NAME = "memcached"
-    KeyType = Union[bytes, bytes]  # Workaround: TypeAlias not in python <= 3.10
 
-    # def __init__(self, serializer=None, *, key_type: CacheKeyType = bytes, **kwargs):
     def __init__(self, serializer=None, **kwargs):
         super().__init__(serializer=serializer or JsonSerializer(), **kwargs)
 
@@ -152,12 +149,8 @@ class MemcachedCache(MemcachedBackend):
     def parse_uri_path(cls, path):
         return {}
 
-    # def build_key(self, key: str, namespace: Optional[str] = None) -> CacheKeyType:
-    #     ns_key = super().build_key(key, namespace=namespace).replace(" ", "_")
-    #     return str.encode(ns_key)
-
-    def build_key(self, key: str, namespace: Optional[str] = None) -> KeyType:
-        ns_key = cast(str, super().build_key(key, namespace=namespace)).replace(" ", "_")
+    def build_key(self, key: str, namespace: Optional[str] = None) -> bytes:
+        ns_key = self._str_build_key(key, namespace).replace(" ", "_")
         return str.encode(ns_key)
 
     def __repr__(self):  # pragma: no cover
--- a/aiocache/backends/memcached.py
+++ b/aiocache/backends/memcached.py
@@ -1,5 +1,5 @@
 import asyncio
-from typing import Optional
+from typing import cast, Optional, Union
 
 import aiomcache
 
@@ -7,6 +7,7 @@ from aiocache.base import BaseCache
 from aiocache.serializers import JsonSerializer
 
 
+# class MemcachedBackend(BaseCache[str]):
 class MemcachedBackend(BaseCache):
     def __init__(self, endpoint="127.0.0.1", port=11211, pool_size=2, **kwargs):
         super().__init__(**kwargs)
@@ -120,6 +121,7 @@ class MemcachedBackend(BaseCache):
         await self.client.close()
 
 
+# class MemcachedCache(Generic[CacheKeyType], MemcachedBackend):
 class MemcachedCache(MemcachedBackend):
     """
     Memcached cache implementation with the following components as defaults:
@@ -140,7 +142,9 @@ class MemcachedCache(MemcachedBackend):
     """
 
     NAME = "memcached"
+    KeyType = Union[bytes, bytes]  # Workaround: TypeAlias not in python <= 3.10
 
+    # def __init__(self, serializer=None, *, key_type: CacheKeyType = bytes, **kwargs):
     def __init__(self, serializer=None, **kwargs):
         super().__init__(serializer=serializer or JsonSerializer(), **kwargs)
 
@@ -148,8 +152,12 @@ class MemcachedCache(MemcachedBackend):
     def parse_uri_path(cls, path):
         return {}
 
-    def build_key(self, key: str, namespace: Optional[str] = None) -> str:
-        ns_key = super().build_key(key, namespace=namespace).replace(" ", "_")
+    # def build_key(self, key: str, namespace: Optional[str] = None) -> CacheKeyType:
+    #     ns_key = super().build_key(key, namespace=namespace).replace(" ", "_")
+    #     return str.encode(ns_key)
+
+    def build_key(self, key: str, namespace: Optional[str] = None) -> KeyType:
+        ns_key = cast(str, super().build_key(key, namespace=namespace)).replace(" ", "_")
         return str.encode(ns_key)
 
     def __repr__(self):  # pragma: no cover
--- a/aiocache/backends/memcached.py
+++ b/aiocache/backends/memcached.py
@@ -1,4 +1,5 @@
 import asyncio
+from typing import Optional
 
 import aiomcache
 
@@ -130,7 +131,7 @@ class MemcachedCache(MemcachedBackend):
     :param serializer: obj derived from :class:`aiocache.serializers.BaseSerializer`.
     :param plugins: list of :class:`aiocache.plugins.BasePlugin` derived classes.
     :param namespace: string to use as default prefix for the key used in all operations of
-        the backend. Default is None
+        the backend. Default is an empty string, "".
     :param timeout: int or float in seconds specifying maximum timeout for the operations to last.
         By default its 5.
     :param endpoint: str with the endpoint to connect to. Default is 127.0.0.1.
@@ -147,7 +148,7 @@ class MemcachedCache(MemcachedBackend):
     def parse_uri_path(cls, path):
         return {}
 
-    def build_key(self, key, namespace=None):
+    def build_key(self, key: str, namespace: Optional[str] = None) -> str:
         ns_key = super().build_key(key, namespace=namespace).replace(" ", "_")
         return str.encode(ns_key)
 
--- a/aiocache/backends/memcached.py
+++ b/aiocache/backends/memcached.py
@@ -120,7 +120,6 @@ class MemcachedBackend(BaseCache[bytes]):
         await self.client.close()
 
 
-# class MemcachedCache(MemcachedBackend):
 class MemcachedCache(MemcachedBackend):
     """
     Memcached cache implementation with the following components as defaults:
--- a/aiocache/backends/memory.py
+++ b/aiocache/backends/memory.py
@@ -1,5 +1,5 @@
 import asyncio
-from typing import Dict
+from typing import Dict, Optional
 
 from aiocache.base import BaseCache
 from aiocache.serializers import NullSerializer
@@ -12,7 +12,10 @@ class SimpleMemoryBackend(BaseCache[str]):
 
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
-        self.build_key = self._str_build_key
+        # Assigning build_key here is clear and avoids the extra stack frame
+        # from nesting the call to _str_build_key() in an override of build_key()
+        # ...but mypy needs the override definition to recognize a concrete class
+        # self.build_key = self._str_build_key  # Simple, but not mypy-friendly
 
         self._cache: Dict[str, object] = {}
         self._handlers: Dict[str, asyncio.TimerHandle] = {}
@@ -133,5 +136,5 @@ class SimpleMemoryCache(SimpleMemoryBackend):
     def parse_uri_path(cls, path):
         return {}
 
-    # def build_key(self, key: str, namespace: Optional[str] = None) -> str:
-    #     return self._str_build_key(key, namespace)
+    def build_key(self, key: str, namespace: Optional[str] = None) -> str:
+        return self._str_build_key(key, namespace)
--- a/aiocache/backends/memory.py
+++ b/aiocache/backends/memory.py
@@ -12,10 +12,6 @@ class SimpleMemoryBackend(BaseCache[str]):
 
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
-        # Assigning build_key here is clear and avoids the extra stack frame
-        # from nesting the call to _str_build_key() in an override of build_key()
-        # ...but mypy needs the override definition to recognize a concrete class
-        # self.build_key = self._str_build_key  # Simple, but not mypy-friendly
 
         self._cache: Dict[str, object] = {}
         self._handlers: Dict[str, asyncio.TimerHandle] = {}
--- a/aiocache/backends/memory.py
+++ b/aiocache/backends/memory.py
@@ -1,10 +1,11 @@
 import asyncio
-from typing import Dict
+from typing import Dict, Union
 
 from aiocache.base import BaseCache
 from aiocache.serializers import NullSerializer
 
 
+# class SimpleMemoryBackend(BaseCache[str]):
 class SimpleMemoryBackend(BaseCache):
     """
     Wrapper around dict operations to use it as a cache backend
@@ -124,6 +125,7 @@ class SimpleMemoryCache(SimpleMemoryBackend):
     """
 
     NAME = "memory"
+    KeyType = Union[str, str]  # Workaround: TypeAlias not in python <= 3.10
 
     def __init__(self, serializer=None, **kwargs):
         super().__init__(serializer=serializer or NullSerializer(), **kwargs)
--- a/aiocache/backends/memory.py
+++ b/aiocache/backends/memory.py
@@ -1,18 +1,18 @@
 import asyncio
-from typing import Dict, Union
+from typing import Dict
 
 from aiocache.base import BaseCache
 from aiocache.serializers import NullSerializer
 
 
-# class SimpleMemoryBackend(BaseCache[str]):
-class SimpleMemoryBackend(BaseCache):
+class SimpleMemoryBackend(BaseCache[str]):
     """
     Wrapper around dict operations to use it as a cache backend
     """
 
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
+        self.build_key = self._str_build_key
 
         self._cache: Dict[str, object] = {}
         self._handlers: Dict[str, asyncio.TimerHandle] = {}
@@ -125,7 +125,6 @@ class SimpleMemoryCache(SimpleMemoryBackend):
     """
 
     NAME = "memory"
-    KeyType = Union[str, str]  # Workaround: TypeAlias not in python <= 3.10
 
     def __init__(self, serializer=None, **kwargs):
         super().__init__(serializer=serializer or NullSerializer(), **kwargs)
@@ -133,3 +132,6 @@ class SimpleMemoryCache(SimpleMemoryBackend):
     @classmethod
     def parse_uri_path(cls, path):
         return {}
+
+    # def build_key(self, key: str, namespace: Optional[str] = None) -> str:
+    #     return self._str_build_key(key, namespace)
--- a/aiocache/backends/memory.py
+++ b/aiocache/backends/memory.py
@@ -118,7 +118,7 @@ class SimpleMemoryCache(SimpleMemoryBackend):
     :param serializer: obj derived from :class:`aiocache.serializers.BaseSerializer`.
     :param plugins: list of :class:`aiocache.plugins.BasePlugin` derived classes.
     :param namespace: string to use as default prefix for the key used in all operations of
-        the backend. Default is None.
+        the backend. Default is an empty string, "".
     :param timeout: int or float in seconds specifying maximum timeout for the operations to last.
         By default its 5.
     """
--- a/aiocache/backends/redis.py
+++ b/aiocache/backends/redis.py
@@ -4,7 +4,7 @@ import warnings
 import redis.asyncio as redis
 from redis.exceptions import ResponseError as IncrbyException
 
-from aiocache.base import BaseCache, _ensure_key
+from aiocache.base import BaseCache
 from aiocache.serializers import JsonSerializer
 
 
@@ -218,14 +218,8 @@ class RedisCache(RedisBackend):
             options["db"] = db
         return options
 
-    def _build_key(self, key, namespace=None):
-        if namespace is not None:
-            return "{}{}{}".format(
-                namespace, ":" if namespace else "", _ensure_key(key))
-        if self.namespace is not None:
-            return "{}{}{}".format(
-                self.namespace, ":" if self.namespace else "", _ensure_key(key))
-        return key
+    def _build_key_default(self, key, namespace=None):
+        return "{}{}{}".format(namespace or "", ":" if namespace else "", key)
 
     def __repr__(self):  # pragma: no cover
         return "RedisCache ({}:{})".format(self.endpoint, self.port)
--- a/aiocache/backends/redis.py
+++ b/aiocache/backends/redis.py
@@ -47,7 +47,10 @@ class RedisBackend(BaseCache[str]):
         **kwargs,
     ):
         super().__init__(**kwargs)
-        self.build_key = self._str_build_key
+        # Assigning build_key here is clear and avoids the extra stack frame
+        # from nesting the call to _str_build_key() in an override of build_key()
+        # ...but mypy needs the override definition to recognize a concrete class
+        # self.build_key = self._str_build_key  # Simple, but not mypy-friendly
         if pool_min_size is not _NOT_SET:
             warnings.warn(
                 "Parameter 'pool_min_size' is deprecated since aiocache 0.12",
@@ -238,5 +241,5 @@ class RedisCache(RedisBackend):
     def __repr__(self):  # pragma: no cover
         return "RedisCache ({}:{})".format(self.endpoint, self.port)
 
-    # def build_key(self, key: str, namespace: Optional[str] = None) -> str:
-    #     return self._str_build_key(key, namespace)
+    def build_key(self, key: str, namespace: Optional[str] = None) -> str:
+        return self._str_build_key(key, namespace)
--- a/aiocache/backends/redis.py
+++ b/aiocache/backends/redis.py
@@ -1,17 +1,20 @@
 import itertools
 import warnings
-from typing import Callable, Optional
+from typing import Any, Callable, Optional, TYPE_CHECKING, Union
 
 import redis.asyncio as redis
 from redis.exceptions import ResponseError as IncrbyException
 
+if TYPE_CHECKING:
+    from aiocache.serializers import BaseSerializer
 from aiocache.base import BaseCache
-from aiocache.serializers import BaseSerializer, JsonSerializer
+from aiocache.serializers import JsonSerializer
 
 
 _NOT_SET = object()
 
 
+# class RedisBackend(BaseCache[str]):
 class RedisBackend(BaseCache):
     RELEASE_SCRIPT = (
         "if redis.call('get',KEYS[1]) == ARGV[1] then"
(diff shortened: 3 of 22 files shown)


aiocache/backends/redis.py
```diff
@@ -218,14 +218,8 @@ def parse_uri_path(cls, path):
...
-    def _build_key(self, key, namespace=None):
-        if namespace is not None:
-            return "{}{}{}".format(
-                namespace, ":" if namespace else "", _ensure_key(key))
-        if self.namespace is not None:
-            return "{}{}{}".format(
-                self.namespace, ":" if self.namespace else "", _ensure_key(key))
-        return key
+    def _build_key_default(self, key, namespace=None):
+        return "{}{}{}".format(namespace or "", ":" if namespace else "", key)
```
Dreamsorcerer, line 222: If we change to always `str`, then this becomes:
```suggestion
        return "{}:{}".format(namespace, key) if namespace else key
```

aiocache/base.py
```diff
@@ -498,12 +489,13 @@ async def close(self, *args, _conn=None, **kwargs):
...
-            return "{}{}".format(namespace, _ensure_key(key))
-        if self.namespace is not None:
-            return "{}{}".format(self.namespace, _ensure_key(key))
-        return key
+    def build_key(self, key, namespace=None):
+        key_name = key.value if isinstance(key, Enum) else key
+        ns = namespace if namespace is not None else self.namespace
+        return self._build_key(key_name, namespace=ns)
+
+    def _build_key_default(self, key, namespace=None):
```
Dreamsorcerer, line 497: I'm still inclined to just pass a lambda as the default value in `__init__()`.

Dreamsorcerer:
Would also be nice to add type annotations to the methods that have been edited/added (and the key_builder parameter), to get some movement in that direction.

pshafer-als:
> Would also be nice to add type annotations to the methods that have been edited/added (and the key_builder parameter), to get some movement in that direction.

Sure.  That should be a simple improvement.

aiocache/base.py
```diff
@@ -498,12 +489,13 @@ async def close(self, *args, _conn=None, **kwargs):
...
+        return "{}{}".format(namespace or "", key)
```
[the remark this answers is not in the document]
pshafer-als: This difference is meaningful at the level above this point (`BaseCache.build_key()`) because `namespace=None` will use the cache's default, whereas `namespace=""` is an override of the cache's default to use no namespace prefix.

I think you're right that at this level (`BaseCache._build_key()`) there is _**probably**_ no difference between `None` and `""`.  However, that does take away that particular freedom from the end user who might care about that distinction in their custom `key_builder`.

My inclination is slightly in favor of keeping this as is, but it's only based on _possibility_ rather than a concrete use case.

aiocache/backends/redis.py
pshafer-als: Right

aiocache/base.py
pshafer-als: Yeah, maybe you're right.  I had thought that this could provide a nice 'template' for end users who write their own key_builder; and it's clear when it gets overridden (like in the redis backend).

But if you think that the benefit is minimal, I'm ok with using the lambda.

aiocache/base.py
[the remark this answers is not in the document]
Dreamsorcerer: Yep, but I can't think of any real cases for it, and ensuring it is always `str` means that the key builders all become simpler code and less likely that a user writes incorrect code (or has to special case `None` even though they will never use it).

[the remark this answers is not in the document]
pshafer-als: That makes sense. It might be the tiniest smidge faster also.

Let's give it a try.

aiocache/base.py
```diff
@@ -109,18 +110,26 @@ class BaseCache:
     NAME: str
 
     def __init__(
-        self, serializer=None, plugins=None, namespace=None, key_builder=None, timeout=5, ttl=None
+        self,
+        serializer: Optional[BaseSerializer] = None,
+        plugins: Optional[List[Any]] = None,  # aiocache.plugins depends on aiocache.base
```
pshafer-als, line 115: Using `Any` because `aiocache.plugins` depends on `aiocache.base`.

I tried [refactoring `API` into its own module](https://github.com/pshafer-als/aiocache/blob/simplify-cache-namespace-refactor-api/aiocache/api.py), but it causes a lot of tests to fail.  At the moment I'm presuming this is `@API.register` writing to the class variables of different objects

aiocache/base.py
```diff
@@ -498,12 +498,13 @@ async def close(self, *args, _conn=None, **kwargs):
...
-        if self.namespace is not None:
-            return "{}{}".format(self.namespace, _ensure_key(key))
-        return key
+    def build_key(self, key: str, namespace: Optional[str] = None) -> str:
+        key_name = key.value if isinstance(key, Enum) else key
+        ns = namespace if namespace is not None else self.namespace
+        return self._build_key(key_name, namespace=ns)
+
+    def _build_key_default(self, key: str, namespace: str = "") -> str:
+        return f"{namespace}{key}"
```
pshafer-als, line 507: Will delete this in the next commit.

pshafer-als:
I added the changes we discussed, and started on the type annotations.  I could continue adding type annotations in these files, but would like to get your feedback before proceeding further.

aiocache/base.py
pshafer-als: > At the moment I'm presuming this is @API.register writing to the class variables of different objects

That wasn't the problem.  It's an 'order of import' problem, where `aiocache.plugins` is attempting to create hooks before `aiocache.base` registers any commands.

Another approach would be to move the base classes into a separate model, so that they can be used for type annotations.  I'll give that a try to see how 'forced' it feels.

pshafer-als: > Another approach would be to move the base classes into a separate module, so that they can be used for type annotations.

Yes, [that approach is cleaner](https://github.com/pshafer-als/aiocache/blob/simplify-cache-namespace-with-types/aiocache/types.py). If this becomes repeatedly necessary as type annotations are added to other parts of the codebase, then we could move other base classes into the 'types' module as needed.

aiocache/base.py
```diff
@@ -109,18 +110,26 @@ class BaseCache:
...
-        self._serializer = None
-        self.serializer = serializer or serializers.StringSerializer()
+        self.namespace: str = namespace
+        self._build_key: Callable[[str, str], str] = key_builder or (
+            lambda key, namespace="": f"{namespace}{key}")
 
-        self._plugins = None
-        self.plugins = plugins or []
+        self._serializer: Optional[BaseSerializer] = None
+        self.serializer: BaseSerializer = serializer or StringSerializer()
```
pshafer-als, line 129: Here and elsewhere, should we replace the base class with a `TypeVar`?

Something like...
```
from typing import TypeVar
SerializerType = TypeVar('SerializerType', bound='BaseSerializer')

self.serializer: SerializerType = serializer or StringSerializer()
```

Dreamsorcerer:
> I added the changes we discussed, and started on the type annotations. I could continue adding type annotations in these files, but would like to get your feedback before proceeding further.

Let's keep the changes to a minimum for this PR, further type annotations can be done in another PR. However, I suspect that completing some of the 1.0 refactoring tasks first will make it easier.

aiocache/backends/memcached.py
```diff
@@ -147,8 +148,8 @@ def __init__(self, serializer=None, **kwargs):
     def parse_uri_path(cls, path):
         return {}
 
-    def _build_key(self, key, namespace=None):
-        ns_key = super()._build_key(key, namespace=namespace).replace(" ", "_")
+    def build_key(self, key: str, namespace: Optional[str] = None) -> str:
```
Dreamsorcerer, line 151: Should probably try to diff the mypy results to see if there are new errors. master currently has 20 errors:
https://github.com/aio-libs/aiocache/actions/runs/4235473010/jobs/7359129847

This must be an error, because the function returns `bytes`.

Maybe we need a `TypeVar` to declare the type here (which can be defined in each subclass).

aiocache/backends/redis.py
```diff
@@ -218,14 +230,8 @@ def parse_uri_path(cls, path):
...
-    def _build_key(self, key, namespace=None):
-        if namespace is not None:
-            return "{}{}{}".format(
-                namespace, ":" if namespace else "", _ensure_key(key))
-        if self.namespace is not None:
-            return "{}{}{}".format(
-                self.namespace, ":" if self.namespace else "", _ensure_key(key))
-        return key
+    def _build_key_default(self, key: str, namespace: str = "") -> str:
+        return f"{namespace}:{key}" if namespace else key
```
Dreamsorcerer, line 234:
```suggestion
```

aiocache/backends/redis.py
```diff
@@ -199,8 +200,19 @@ class RedisCache(RedisBackend):
...
+        serializer: Optional[BaseSerializer] = None,
+        namespace: str = "",
+        key_builder: Optional[Callable[[str, str], str]] = None,
+        **kwargs,
+    ):
+        super().__init__(
+            serializer=serializer or JsonSerializer(),
+            namespace=namespace,
+            key_builder=key_builder or (
+                lambda key, namespace="": f"{namespace}:{key}" if namespace else key),
```
Dreamsorcerer, line 214: Default is unneeded, we always pass namespace in build_key().
```suggestion
                lambda key, namespace: f"{namespace}:{key}" if namespace else key),
```

aiocache/base.py
Dreamsorcerer: For now, just import it with `if TYPE_CHECKING:`. Further refactoring can be considered in another PR if you still feel it's appropriate.

aiocache/base.py
```diff
@@ -109,18 +110,26 @@ class BaseCache:
...
+        key_builder: Optional[Callable[[str, str], str]] = None,
+        timeout: Optional[float] = 5,
+        ttl: Optional[float] = None,
     ):
-        self.timeout = float(timeout) if timeout is not None else timeout
-        self.namespace = namespace
-        self.ttl = float(ttl) if ttl is not None else ttl
-        self.build_key = key_builder or self._build_key
+        self.timeout: Optional[float] = float(timeout) if timeout is not None else None
+        self.ttl: Optional[float] = float(ttl) if ttl is not None else None
```
Dreamsorcerer, line 122:
```suggestion
        self.timeout = float(timeout) if timeout is not None else None
        self.ttl = float(ttl) if ttl is not None else None
```

Double check with `reveal_type`, but mypy should infer the types for these attributes.

aiocache/base.py
```diff
@@ -109,18 +110,26 @@ class BaseCache:
...
 
-        self._serializer = None
-        self.serializer = serializer or serializers.StringSerializer()
+        self.namespace: str = namespace
+        self._build_key: Callable[[str, str], str] = key_builder or (
```
Dreamsorcerer, line 125:
```suggestion
        self._build_key = key_builder or (
```
Hopefully this one too.

aiocache/base.py
```diff
@@ -109,18 +110,26 @@ class BaseCache:
...
+            lambda key, namespace="": f"{namespace}{key}")
```
Dreamsorcerer, line 126:
```suggestion
            lambda key, namespace: f"{namespace}{key}")
```

aiocache/base.py
Dreamsorcerer: I'm not sure what benefit that brings. But, leave it to a later PR if you think there is a benefit.

aiocache/base.py
```diff
@@ -109,18 +110,26 @@ class BaseCache:
...
 
-        self._plugins = None
-        self.plugins = plugins or []
+        self._serializer: Optional[BaseSerializer] = None
+        self.serializer: BaseSerializer = serializer or StringSerializer()
+
+        self._plugins: List[Any] = None
+        self.plugins: List[Any] = plugins or []
```
Dreamsorcerer, line 132:
```suggestion
        self._serializer = None
        self.serializer = serializer or StringSerializer()

        self._plugins: Optional[List[BasePluign]] = None
        self.plugins = plugins or []
```

Not sure what I'm missing, but `_plugins` doesn't make much sense. I'm not sure why it's initialised to `None` or where it later gets set to something else.

Dreamsorcerer: `plugins` is also a property method, so that's broken too... Maybe someone meant to do `_plugins = plugins or []` and then use the property method?

aiocache/base.py
```diff
@@ -418,8 +419,7 @@ async def expire(self, key, ttl, namespace=None, _conn=None):
         :raises: :class:`asyncio.TimeoutError` if it lasts more than self.timeout
         """
         start = time.monotonic()
-        ns = namespace if namespace is not None else self.namespace
-        ns_key = self.build_key(key, namespace=ns)
+        ns_key = self.build_key(key, namespace=namespace)
```
Dreamsorcerer, line 422: I'd be tempted to remove the kwarg from all these, rather than repeating the name.

pshafer-als:
Later today I'll clean this up per your recommendations.  After that I'm thinking it might make sense in this PR to replace `namespace=None` with `namespace=""` in the call signature for all the BaseCache member functions.

e.g.
https://github.com/aio-libs/aiocache/blob/db9426196d0a4b55a78411c8b96949af645a557b/aiocache/base.py#L154

What do you think?

aiocache/base.py
```diff
@@ -498,12 +498,13 @@ async def close(self, *args, _conn=None, **kwargs):
...
-    def _build_key(self, key, namespace=None):
-        if namespace is not None:
-            return "{}{}".format(namespace, _ensure_key(key))
-        if self.namespace is not None:
-            return "{}{}".format(self.namespace, _ensure_key(key))
-        return key
+    def build_key(self, key: str, namespace: Optional[str] = None) -> str:
+        key_name = key.value if isinstance(key, Enum) else key
+        ns = namespace if namespace is not None else self.namespace
+        return self._build_key(key_name, namespace=ns)
```
Dreamsorcerer, line 504:
```suggestion
        return self._build_key(key_name, ns)
```
Our typing guarantees 2 positional arguments, it doesn't guarantee the keyword name.

tests/ut/test_base.py
```diff
@@ -218,16 +218,15 @@ def test_alt_build_key(self):
     def alt_base_cache(self, init_namespace="test"):
         """Custom key_builder for cache"""
         def build_key(key, namespace=None):
-            ns = namespace if namespace is not None else ""
             sep = ":" if namespace else ""
-            return f"{ns}{sep}{_ensure_key(key)}"
+            return f'{namespace or ""}{sep}{ensure_key(key)}'
```
Dreamsorcerer, line 222:
```suggestion
            return f'{namespace}{sep}{ensure_key(key)}'
```

Dreamsorcerer:
> Later today I'll clean this up per your recommendations. After that I'm thinking it might make sense in this PR to replace `namespace=None` with `namespace=""` in the call signature for all the BaseCache member functions.
> 
> e.g.
> 
> https://github.com/aio-libs/aiocache/blob/db9426196d0a4b55a78411c8b96949af645a557b/aiocache/base.py#L154
> 
> What do you think?

Wouldn't that break the default namespace behaviour? i.e. It needs to pass `None` to `build_key()` to use the default namespace.

aiocache/base.py
```diff
@@ -498,12 +498,13 @@ async def close(self, *args, _conn=None, **kwargs):
...
 
-    def _build_key(self, key, namespace=None):
-        if namespace is not None:
-            return "{}{}".format(namespace, _ensure_key(key))
-        if self.namespace is not None:
-            return "{}{}".format(self.namespace, _ensure_key(key))
-        return key
+    def build_key(self, key: str, namespace: Optional[str] = None) -> str:
+        key_name = key.value if isinstance(key, Enum) else key
+        ns = namespace if namespace is not None else self.namespace
```
Dreamsorcerer, line 503:
```suggestion
        ns = self.namespace if namespace is None else namespace
```
I'd always try to do the positive test.

pshafer-als:
> > After that I'm thinking it might make sense in this PR to replace `namespace=None` with `namespace=""` in the call signature for all the BaseCache member functions.
>
> Wouldn't that break the default namespace behaviour? i.e. It needs to pass `None` to `build_key()` to use the default namespace.

Er, yeah it would.  I came back here to say the same thing, but you beat me to it.  :)

aiocache/backends/memcached.py
Dreamsorcerer: Or 12 errors after installing all missing libraries:
https://github.com/aio-libs/aiocache/actions/runs/4235831591/jobs/7359931126

aiocache/base.py
pshafer-als: That's to indicate that the type can be a subclass of the bound type; so `StringSerializer` is not expected to be coerced to `BaseSerializer`, for example.

> But, leave it to a later PR if you think there is a benefit.

Agreed.

aiocache/backends/memcached.py
pshafer-als: To use a `TypeVar` in the return type, we'll also need either:
1. The same `TypeVar` to annotate on the function parameters.  We could add a third, otherwise unused, parameter for this purpose, but that seems pretty kludged.
2. Derive `BaseCache` from `Generic[]`.  That also seems heavy handed, just for type checking.

Using a simple `Union[str, bytes]` will work as a return type for all existing cases. This could be done with a type alias at the module level. Trying to add the alias within `BaseCache` and its subclasses requires the `TypeAlias` annotation, which is only available for python >= 3.10.

I'll see whether this approach plays well with mypy.

aiocache/base.py
pshafer-als: Mypy is fine without this annotation, but it might be useful for documenting the code?  This is because lambdas can't have type annotations.

aiocache/base.py
pshafer-als: You're right.  This is strange.  I'll reduce these two initialization assignments into one, for `_serializer` and for `_plugins`.

`self.serializer` and `self._serializer` are used interchangeably throughout this module.  `self.serializer` would be appropriate in all cases.

---
Although the getter/setter don't do anything useful now, they do provide a mechanism for subclasses to do any customizations they might need.  I'd like to defer removing them to a future PR.

aiocache/base.py
pshafer-als: Good point.

aiocache/base.py
```diff
@@ -109,18 +110,26 @@ class BaseCache:
     NAME: str
 
     def __init__(
-        self, serializer=None, plugins=None, namespace=None, key_builder=None, timeout=5, ttl=None
+        self,
+        serializer: Optional[BaseSerializer] = None,
+        plugins: Optional[List[Any]] = None,  # aiocache.plugins depends on aiocache.base
+        namespace: str = "",
```
pshafer-als, line 116: This breaks `build_key()` if `None` is passed as an argument here. 

Either we allow `None` here and convert it to `""` inside `__init__()`, or we should add a test that expects to fail when `None` is passed here.

Which would you recommend?

aiocache/base.py
```diff
@@ -87,6 +91,7 @@ async def _plugins(self, *args, **kwargs):
         return _plugins
 
 
+# class BaseCache(Generic[CacheKeyType]):
```
pshafer-als, line 94: Here and elsewhere in comments are how we might implement typing of the `build_key()` return type if we decide using Generics would be more useful or robust.

pshafer-als: It probably means also updating tests that check the call signature of `__init()__`, and maybe a few others.

pshafer-als:
@Dreamsorcerer I'll need to pull in the latest dependency bumps, and update at least one failing test still.  I'm pausing now to share the current state of the PR for your feedback.

aiocache/backends/memcached.py
```diff
@@ -139,16 +142,22 @@ class MemcachedCache(MemcachedBackend):
...
         return {}
 
-    def _build_key(self, key, namespace=None):
-        ns_key = super()._build_key(key, namespace=namespace).replace(" ", "_")
+    # def build_key(self, key: str, namespace: Optional[str] = None) -> CacheKeyType:
+    #     ns_key = super().build_key(key, namespace=namespace).replace(" ", "_")
+    #     return str.encode(ns_key)
+
+    def build_key(self, key: str, namespace: Optional[str] = None) -> KeyType:
+        ns_key = cast(str, super().build_key(key, namespace=namespace)).replace(" ", "_")
```
pshafer-als, line 160: An alternative to `cast` could simply be `# type: ignore`

aiocache/backends/memcached.py
Dreamsorcerer: Yes, I meant with a `Generic`, the functions for `_get()` etc. would all need to use the same TypeVar for the key parameter. This should then type check correctly when you pass the key to memcached which expects bytes (MemcacheBackend would subclass `BaseCache[bytes]`).

aiocache/base.py
Dreamsorcerer: A runtime check? Sure, that's fine.

I'd vote for not allowing `None` though and keeping the code simpler.

aiocache/base.py
Dreamsorcerer: I don't think anything needs to be changed to `__init__()`.

aiocache/backends/memcached.py
Dreamsorcerer: Hmm, yes, that's slightly awkward.

Maybe a type safe option would be to make the `BaseCache` define it as an abstract method, and provide this default method under another name (like _str_build_key() or something)?

pshafer-als: > Maybe a type safe option would be to make the `BaseCache` define it as an abstract method, and provide this default method under another name (like `_str_build_key()` or something)?

I think you're right.  I've been trying to make this work by overloading `build_key()` with `@singledispatchmethod` to no avail.  For that I would need to implement some factory function that returns a `BaseCache` with the appropriate return type for `build_key()`.  Implementing this private helper method seems much simpler.

aiocache/base.py
pshafer-als: I was referring to this potential change [_below_], which was intended as a convenience...providing the default`build_key()` return type of `str` so that it could be inferred by mypy, even when not explicitly created with `BaseCache[str]`.

https://github.com/aio-libs/aiocache/blob/84ce34c07b1e2961583c9f3beeb2c146da2ba3d4/aiocache/base.py#L118-L126
[_although the last line should have been `key_type:Type[CacheKeyType] = str`_]

But, unfortunately, it's not (yet) possible to supply a [default type for the `TypeVar`](https://github.com/python/mypy/issues/4236). Oh well.

Dreamsorcerer: You can use overloads to achieve that, but I don't think it's relevant. There should be no parameter here, all subclasses should explicitly define the type they use.

aiocache/base.py
pshafer-als: Generally I like to call with keyword args if I think the function signature is likely to change (and of course if some kwargs with defaults are skipped).

But in this case I think it's fine use the positional arguments as you've suggested.  The argument name at least conveys the intent.

aiocache/base.py
pshafer-als: > Frankly, you could just define this lambda as the parameter default and disallow `None` completely.

I like this suggestion, but mypy does not. :(

aiocache/backends/memcached.py
pshafer-als: `build_key()` is now an abstract method on `BaseCache`; and overloaded/defined for the derived classes.

aiocache/base.py
pshafer-als: The extra `__init__()` parameter for implicit typing idea has been scrapped.

aiocache/base.py
pshafer-als: For now, have omitted a runtime check.  Passing `namespace=None` does not cause an error; instead it prepends the "None" namespace to the key.

This is probably unexpected behavior, but doesn't cause any real problem.  I could add a runtime warning to alert the user that `""` (or no argument) is the expected value for a cache constructor.

aiocache/base.py
Dreamsorcerer: type: ignore it if it's a bug in mypy.

I know there's this issue around lambdas already: https://github.com/python/mypy/issues/9590

aiocache/backends/memcached.py
```diff
@@ -119,6 +120,7 @@ async def _close(self, *args, _conn=None, **kwargs):
         await self.client.close()
 
 
+# class MemcachedCache(MemcachedBackend):
```
Dreamsorcerer, line 123:
```suggestion
```

aiocache/base.py
pshafer-als: It turns out this was my error.  I think I must have missed a comma at the end of the line or something mundane like that.  Correcting it now.  Will push in the next round of updates for this PR.

aiocache/backends/memory.py
```diff
@@ -1,17 +1,21 @@
...
     """
     Wrapper around dict operations to use it as a cache backend
     """
 
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
+        # Assigning build_key here is clear and avoids the extra stack frame
+        # from nesting the call to _str_build_key() in an override of build_key()
+        # ...but mypy needs the override definition to recognize a concrete class
+        # self.build_key = self._str_build_key  # Simple, but not mypy-friendly
```
Dreamsorcerer, line 18:
```suggestion
```

Not just mypy, I'd expect it to fail at runtime too if it's abstract.

aiocache/backends/memory.py
```diff
@@ -131,3 +135,6 @@ def __init__(self, serializer=None, **kwargs):
     @classmethod
     def parse_uri_path(cls, path):
         return {}
+
+    def build_key(self, key: str, namespace: Optional[str] = None) -> str:
+        return self._str_build_key(key, namespace)
```
Dreamsorcerer, line 140: You might, however, be able to just do:
`build_key = _str_build_key`
Rather than trying to set it in the `__init__()` method.

aiocache/base.py
```diff
@@ -3,16 +3,21 @@
...
+if TYPE_CHECKING:
+    from aiocache.plugins import BasePlugin
+    from aiocache.serializers import BaseSerializer
+from aiocache.serializers import StringSerializer
 
 
 logger = logging.getLogger(__name__)
 
 SENTINEL = object()
+CacheKeyType = TypeVar("CacheKeyType", bound=Union[str, bytes])
```
Dreamsorcerer, line 20: I think it's totally up to the class what type they use for cache keys, so I'd leave this unbounded.

Maybe some special cache uses `int` after producing some kind of hash?

pshafer-als: OK, now that we use the `_str_build_key()` helper, this bound is more restrictive than helpful.

aiocache/factory.py
```diff
@@ -172,12 +172,12 @@ def get(self, alias: str) -> object:
...
         return cache
 
-    def create(self, alias: str, **kwargs):
+    def create(self, alias, **kwargs):
         """Create a new cache.
 
         You can use kwargs to pass extra parameters to configure the cache.
 
-        :param alias: alias to pull configuration from
+        :param alias (str): alias to pull configuration from
```
Dreamsorcerer, line 180: Not sure why this file has been changed, seems unrelated.

pshafer-als: It was a mypy error.  Should I revert it and submit a separate PR?

tests/performance/test_footprint.py
```diff
@@ -1,13 +1,15 @@
...
+from typing import Any, AsyncGenerator
 
 import aiomcache
 import pytest
 import redis.asyncio as redis
 
 
 @pytest.fixture
-async def redis_client() -> redis.Redis:
+async def redis_client() -> "AsyncGenerator[redis.Redis[Any], None]":
```
Dreamsorcerer, line 11:
```suggestion
async def redis_client() -> "AsyncIterator[redis.Redis[Any]]":
```

tests/performance/test_footprint.py
```diff
@@ -1,13 +1,15 @@
...
+    r: "redis.Redis[Any]"
```
Dreamsorcerer, line 12: Not clear why this is needed.

pshafer-als: It was a mypy error.

tests/ut/conftest.py
```diff
@@ -29,27 +29,39 @@ def reset_caches():
...
             stack.enter_context(patch.object(cache, f, autospec=True))
         stack.enter_context(patch.object(cache, "_serializer", autospec=True))
+        stack.enter_context(patch.object(cache, "build_key", cache._str_build_key))
         yield cache
 
 
+@pytest.fixture
+def generic_base_cache():
+    # TODO: Is there need for a separate BaseCache[bytes] fixture?
+    return BaseCache[str]()
```
Dreamsorcerer, line 56: Also, task for another PR. BaseCache should be ABC, which means it then won't be able to be instantiated directly. This should probably then just use the memory cache for testing or something..

tests/ut/test_decorators.py
```diff
@@ -60,7 +60,7 @@ def test_fails_at_instantiation(self):
         with pytest.raises(TypeError):
 
             @cached(wrong_param=1)
-            async def fn() -> None:
+            async def fn():
```
Dreamsorcerer, line 63: This file looks unrelated again.

aiocache/backends/memory.py
pshafer-als: I'm having no luck with that approach.

aiocache/backends/memory.py
Dreamsorcerer: That's because it's not actually an ABC, which it should be (later PR though).

aiocache/factory.py
Dreamsorcerer: Ignore the existing 12 errors:
https://github.com/aio-libs/aiocache/actions/runs/4235831591/jobs/7359931126

They can be dealt with separately. I think we just want to remove this entire file, as I don't see what benefit this part adds (but makes the code unnecessarily complex and fail type checking).

pshafer-als: OK, I'll remove the extraneous mypy fixes from this PR.

tests/ut/test_decorators.py
pshafer-als: Reverted these changes

tests/ut/conftest.py
pshafer-als: Sounds good.

tests/performance/test_footprint.py
pshafer-als: Reverted these changes

tests/performance/test_footprint.py
pshafer-als: Done (and then reverted per subsequent comment)

aiocache/base.py
pshafer-als: Will leave any getter/setter changes to a subsequent PR.

Dreamsorcerer:
If you want to work on any more, I think #676 and #625 are probably good next issues.

Dreamsorcerer:
#677 as well. Should be easier to get some typing working then.

PR #671, opened 2023-02-21
Run mypy

Dreamsorcerer:


elecnix:
What’s involved here? What’s the goal?

Fixing all the problems reported by mypy?

PR #682, opened 2023-03-03
Only include aiocache in packages

### LINKED ISSUE #681: Tests folder being imported when installing version 0.12.0
When importing aiocache, it is adding it's tests folder to the virtual env. This is overriding tests in any python solution. For example, the solution:

- src
  - my_module.py
- tests
  - test_my_module.py
  - test_helpers.py

when importing test_helpers from test_my_module:

```python
from tests.test_helpers import helper
```

The aiocache tests module is resolved, not the tests in the solution.

Dreamsorcerer:
Fixes #681.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -29,7 +29,7 @@ setup(
         "Programming Language :: Python :: 3.11",
         "Framework :: AsyncIO",
     ],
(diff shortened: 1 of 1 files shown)


tomsoliver:
Thanks for the quick turn around 🙏

Issue #676, opened 2023-02-27
BaseCache should be ABC


Dreamsorcerer:
BaseCache should be turned into an ABC with abstractmethods.

pshafer-als:
I'll give this a go.

Dreamsorcerer:
#683

PR #683, opened 2023-03-04
Make BaseCache an ABC

pshafer-als:
## What do these changes do?

`BaseCache` is now an abstract base class. Attempting to instantiate a `BaseCache` object raises a `TypeError`.

Generally not, because users have always been expected to create cache objects using a _subclass_ of `BaseCache`.

Now a `TypeError` is raised at the point of attempted cache creation, if the user attempts to instantiate a `BaseCache` object. Previously a `NotImplementedError` was raised at the point where an abstract method was called.

## Related issue number

#676 

## Checklist

- [ X] I think the code is well written
- [ X] Unit tests for the changes exist
- [N/A ] Documentation reflects the changes
- [ X] Add a new news fragment into the `CHANGES` folder

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,6 +16,7 @@ There are a number of backwards-incompatible changes. These points should help w
 * The ``key`` parameter has been removed from the ``cached`` decorator. The behaviour can be easily reimplemented with ``key_builder=lambda *a, **kw: "foo"``
 * When using the ``key_builder`` parameter in ``@multicached``, the function will now return the original, unmodified keys, only using the transformed keys in the cache (this has always been the documented behaviour, but not the implemented behaviour).
 * ``BaseSerializer`` is now an ``ABC``, so cannot be instantiated directly.
+* ``BaseCache`` is now an ``ABC``, so cannot be instantiated directly.
 * If subclassing ``BaseCache`` to implement a custom backend:
 
   * The cache key type used by the backend must now be specified when inheriting (e.g. ``BaseCache[str]`` typically).
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -15,8 +15,7 @@ There are a number of backwards-incompatible changes. These points should help w
 * The ``key_builder`` parameter for caches now expects a callback which accepts 2 strings and returns a string in all cache implementations, making the builders simpler and interchangeable.
 * The ``key`` parameter has been removed from the ``cached`` decorator. The behaviour can be easily reimplemented with ``key_builder=lambda *a, **kw: "foo"``
 * When using the ``key_builder`` parameter in ``@multicached``, the function will now return the original, unmodified keys, only using the transformed keys in the cache (this has always been the documented behaviour, but not the implemented behaviour).
-* ``BaseSerializer`` is now an ``ABC``, so cannot be instantiated directly.
-* ``BaseCache`` is now an ``ABC``, so cannot be instantiated directly.
+* ``BaseCache`` and ``BaseSerializer`` are now ``ABC``s, so cannot be instantiated directly.
 * If subclassing ``BaseCache`` to implement a custom backend:
 
   * The cache key type used by the backend must now be specified when inheriting (e.g. ``BaseCache[str]`` typically).
--- a/aiocache/base.py
+++ b/aiocache/base.py
@@ -509,9 +509,8 @@ class BaseCache(Generic[CacheKeyType], ABC):
         logger.debug("CLOSE (%.4f)s", time.monotonic() - start)
         return ret
 
-    @abstractmethod
     async def _close(self, *args, **kwargs):
-        raise NotImplementedError()
+        pass
 
     @abstractmethod
     def build_key(self, key: str, namespace: Optional[str] = None) -> CacheKeyType:
--- a/aiocache/base.py
+++ b/aiocache/base.py
@@ -511,7 +511,7 @@ class BaseCache(Generic[CacheKeyType], ABC):
 
     @abstractmethod
     async def _close(self, *args, **kwargs):
-        pass
+        raise NotImplementedError()
 
     @abstractmethod
     def build_key(self, key: str, namespace: Optional[str] = None) -> CacheKeyType:
--- a/aiocache/base.py
+++ b/aiocache/base.py
@@ -3,7 +3,7 @@ import functools
 import logging
 import os
 import time
-from abc import abstractmethod
+from abc import ABC, abstractmethod
 from enum import Enum
 from types import TracebackType
 from typing import Callable, Generic, List, Optional, Set, TYPE_CHECKING, Type, TypeVar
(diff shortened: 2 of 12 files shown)


aiocache/base.py
```diff
@@ -496,6 +509,7 @@ async def close(self, *args, _conn=None, **kwargs):
         logger.debug("CLOSE (%.4f)s", time.monotonic() - start)
         return ret
 
+    @abstractmethod
     async def _close(self, *args, **kwargs):
         pass
```
pshafer-als, line 514:
```suggestion
        raise NotImplementedError()
```

tests/ut/backends/test_memcached.py
```diff
@@ -1,14 +1,21 @@
...
 from aiocache.serializers import JsonSerializer
 from ...utils import Keys, ensure_key
 
 
+class MemcachedBackend(_MemcachedBackend):
+    """Define a concrete version of the abstract class, MemcachedBackend"""
+    def build_key(self, key: str, namespace: Optional[str] = None) -> str:
+        return super().build_key(key, namespace)
+
+
```
Dreamsorcerer, line 18: I think it probably makes more sense to just import MemcachedCache?

I suspect they should also be merged into a single class, I really don't see why they are split into 2 classes to start with..

tests/ut/backends/test_memory.py
```diff
@@ -1,14 +1,23 @@
...
+)
 from aiocache.base import BaseCache
 from aiocache.serializers import NullSerializer
 from ...utils import Keys
 
 
+class SimpleMemoryBackend(_SimpleMemoryBackend):
+    """Define a concrete version of the abstract class, SimpleMemoryBackend"""
+    def build_key(self, key: str, namespace: Optional[str] = None) -> str:
+        return super().build_key(key, namespace)
```
Dreamsorcerer, line 18: Same

tests/ut/backends/test_redis.py
```diff
@@ -1,15 +1,22 @@
...
+from aiocache.backends.redis import RedisBackend as _RedisBackend, RedisCache
 from aiocache.base import BaseCache
 from aiocache.serializers import JsonSerializer
 from ...utils import Keys, ensure_key
 
 
+class RedisBackend(_RedisBackend):
+    """Define a concrete version of the abstract class, RedisBackend"""
+    def build_key(self, key: str, namespace: Optional[str] = None) -> str:
+        return super().build_key(key, namespace)
```
Dreamsorcerer, line 17: And once more.

aiocache/base.py
```diff
@@ -496,8 +509,9 @@ async def close(self, *args, _conn=None, **kwargs):
         logger.debug("CLOSE (%.4f)s", time.monotonic() - start)
         return ret
 
+    @abstractmethod
     async def _close(self, *args, **kwargs):
-        pass
+        raise NotImplementedError()
```
Dreamsorcerer, line 514: Maybe this one shouldn't be abstract?

Many backends may not need to close, so a default `pass` is probably good. Whereas the other methods marked as abstract are actually needed to provide a fully functioning cache.

CHANGES.rst
```diff
@@ -16,6 +16,7 @@ There are a number of backwards-incompatible changes. These points should help w
 * The ``key`` parameter has been removed from the ``cached`` decorator. The behaviour can be easily reimplemented with ``key_builder=lambda *a, **kw: "foo"``
 * When using the ``key_builder`` parameter in ``@multicached``, the function will now return the original, unmodified keys, only using the transformed keys in the cache (this has always been the documented behaviour, but not the implemented behaviour).
 * ``BaseSerializer`` is now an ``ABC``, so cannot be instantiated directly.
+* ``BaseCache`` is now an ``ABC``, so cannot be instantiated directly.
```
Dreamsorcerer, line 19:
```suggestion
* ``BaseCache`` and ``BaseSerializer`` are now ``ABC``s, so cannot be instantiated directly.
```

tests/ut/backends/test_memcached.py
pshafer-als: > I suspect they should also be merged into a single class, I really don't see why they are split into 2 classes to start with..

You might be right.  I too don't quite see the value in having them separate, but I would want to spend some more time assessing that.

As a quick fix, I'll move `build_key()` to the `...Backend` classes, and submit an issue for combing the pair of derived classes for each backend implementation.

pshafer-als: > As a quick fix, I'll move `build_key()` to the `...Backend` classes, and submit an issue for combing the pair of derived classes for each backend implementation.

Submitted https://github.com/aio-libs/aiocache/issues/684

tests/utils.py
```diff
@@ -9,8 +12,65 @@ class Keys(str, Enum):
...
-def ensure_key(key):
+def ensure_key(key: Union[str, Enum]) -> str:
     if isinstance(key, Enum):
         return key.value
     else:
         return key
+
+
+class AbstractBaseCache(BaseCache[str]):
+    """BaseCache that can be mocked for NotImplementedError tests"""
```
Dreamsorcerer, line 23: I feel like these 2 classes are rather awkward, but I'll merge it for now.

I don't think there's any value in testing the logic of abstract methods, and I'm wondering if the other tests would still be testing the same thing if we just use the memory cache instead..

Dreamsorcerer: Also introduces a new mypy error:
https://github.com/aio-libs/aiocache/actions/runs/4333749520/jobs/7567135931

Curiously, only on the first method, though it looks like it should happen on all of them. Maybe because it doesn't have an await...

pshafer-als: > Also introduces a new mypy error: https://github.com/aio-libs/aiocache/actions/runs/4333749520/jobs/7567135931

Because this class exists purely for tests, to allow mocks without throwing ABC `TypeError`s, I think adding `# type: ignore[safe-super]` would be appropriate here.

Dreamsorcerer: I realise that, it's just a side note to the first comment about whether we can just remove them if they are not testing anything useful.

pshafer-als: > I feel like these 2 classes are rather awkward, but I'll merge it for now.
> 

Yeah, it's a bit awkward because the tests aren't being performed **_directly_** on `BaseCache`, and because of how many abstract methods need to be defined on the derived class.  However the definitions of `AbstractBaseCache` and `ConcreteBaseCache` are trivial and transparent, so at least no new complexity is added.


> I don't think there's any value in testing the logic of abstract methods, and I'm wondering if the other tests would still be testing the same thing if we just use the memory cache instead.. If so, we could then drop both of these classes.

On the contrary, I really like how the logic of what's defined in `BaseCache` gets tested independently with the (abstract) base class.  

The derived classes used for the backend implementations use more resources and add enough "side effects" that proper testing of the underlying logic would require repeating these tests with all 3 backends.  And even then, it would perhaps be an open question of whether the `BaseCache` functionality extends to new backends.

Keeping these tests of `BaseCache` separate from tests of the backends seems much cleaner and more robust. Previously this was a bit simpler because `BaseCache` was not formally abstract--it had unimplemented methods intended to be override, but instantiating `BaseCache` did not raise an exception.

Well, in any case, I'm open to finding a cleaner test solution, but I favor keeping the `BaseCache` tests separate from the backend tests of the same methods.

PR #685, opened 2023-03-05
Cleanup BaseCache unit tests with an alternate build_key and default namespace

pshafer-als:
## What do these changes do?

This is a minor cleanup for a set of unit tests of `BaseCache` that use an alternate build_key for the cache and that rely on using the default namespace of the cache. The tests remain the same. This is simply a code clean up that better organizes and documents helper methods for these tests, which no longer litter the namespace of `TestBaseCache`.

No.

## Related issue number

N/A

## Checklist

- [ N/A] Unit tests for the changes exist
- [N/A ] Add a new news fragment into the `CHANGES` folder

### DIFF
--- a/tests/ut/test_base.py
+++ b/tests/ut/test_base.py
@@ -260,64 +260,77 @@ class TestBaseCache:
         cache = ConcreteBaseCache(key_builder=self.alt_build_key, namespace=namespace)
 
         # Verify that private members are called with the correct ns_key
-        await self._assert_add__alt_build_key_default_namespace(cache, expected)
-        await self._assert_get__alt_build_key_default_namespace(cache, expected)
-        await self._assert_multi_get__alt_build_key_default_namespace(cache, expected)
-        await self._assert_set__alt_build_key_default_namespace(cache, expected)
-        await self._assert_multi_set__alt_build_key_default_namespace(cache, expected)
-        await self._assert_exists__alt_build_key_default_namespace(cache, expected)
-        await self._assert_increment__alt_build_key_default_namespace(cache, expected)
-        await self._assert_delete__alt_build_key_default_namespace(cache, expected)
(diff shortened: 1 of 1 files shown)


tests/ut/test_base.py
```diff
@@ -260,64 +260,77 @@ async def test_alt_build_key_default_namespace(self, namespace, expected):
...
+        await BaseCacheBuildKeyMethods.assert_get(cache, expected)
+        await BaseCacheBuildKeyMethods.assert_multi_get(cache, expected)
+        await BaseCacheBuildKeyMethods.assert_set(cache, expected)
+        await BaseCacheBuildKeyMethods.assert_multi_set(cache, expected)
+        await BaseCacheBuildKeyMethods.assert_exists(cache, expected)
+        await BaseCacheBuildKeyMethods.assert_increment(cache, expected)
+        await BaseCacheBuildKeyMethods.assert_delete(cache, expected)
+        await BaseCacheBuildKeyMethods.assert_expire(cache, expected)
+
+    class BaseCacheBuildKeyMethods:
```
Dreamsorcerer, line 274: I don't see what benefit a nested class adds here. Sticking to methods on the existing class (or just global functions as it doesn't appear to need a reference to the instance anyway) seems better.

pshafer-als: ok, fair enough. 

I find grouping these related methods into a class cleaner, but I think it comes down to individual preferences. Let's skip this PR then.  :)

Dreamsorcerer: Alright, I wouldn't be opposed to grouping into a module or something. But, not really important here.

I tend to question any use of @staticmethod, as per https://wemake-python-styleguide.readthedocs.io/en/latest/pages/usage/violations/oop.html#wemake_python_styleguide.violations.oop.StaticMethodViolation

PR #691, opened 2023-03-15
add ssl support

cancan101:
## Checklist

### DIFF
--- a/aiocache/backends/redis.py
+++ b/aiocache/backends/redis.py
@@ -42,6 +42,7 @@ class RedisBackend(BaseCache):
         pool_max_size=None,
         create_connection_timeout=None,
         ssl=False,
+        connection_pool_class=redis.ConnectionPool,
         connection_pool_kwargs=None,
         **kwargs,
     ):
@@ -72,7 +73,7 @@ class RedisBackend(BaseCache):
         # (decode_responses=False), because some of the values are saved as
         # bytes directly, like pickle serialized values, which may raise an
         # exception when decoded with 'utf-8'.
-        connection_pool = redis.ConnectionPool(
+        connection_pool = connection_pool_class(
             host=self.endpoint, port=self.port, db=self.db,
             password=self.password, decode_responses=False,
             socket_connect_timeout=self.create_connection_timeout,
--- a/aiocache/backends/redis.py
+++ b/aiocache/backends/redis.py
@@ -42,7 +42,7 @@ class RedisBackend(BaseCache):
         pool_max_size=None,
         create_connection_timeout=None,
         ssl=False,
-        connection_pool_class=redis.ConnectionPool,
+        connection_pool_class=None,
         connection_pool_kwargs=None,
         **kwargs,
     ):
@@ -73,6 +73,7 @@ class RedisBackend(BaseCache):
         # (decode_responses=False), because some of the values are saved as
         # bytes directly, like pickle serialized values, which may raise an
         # exception when decoded with 'utf-8'.
+        connection_pool_class = connection_pool_class or redis.ConnectionPool
         connection_pool = connection_pool_class(
             host=self.endpoint, port=self.port, db=self.db,
             password=self.password, decode_responses=False,
--- a/aiocache/backends/redis.py
+++ b/aiocache/backends/redis.py
@@ -41,6 +41,8 @@ class RedisBackend(BaseCache):
         pool_min_size=_NOT_SET,
         pool_max_size=None,
         create_connection_timeout=None,
+        ssl=False,
+        connection_pool_kwargs=None,
         **kwargs,
     ):
         super().__init__(**kwargs)
@@ -60,15 +62,24 @@ class RedisBackend(BaseCache):
             float(create_connection_timeout) if create_connection_timeout else None
         )
 
+        connection_pool_kwargs = connection_pool_kwargs or {}
+
+        if ssl:
+            connection_pool_kwargs["connection_class"] = redis.SSLConnection
+
         # NOTE: decoding can't be controlled on API level after switching to
         # redis, we need to disable decoding on global/connection level
         # (decode_responses=False), because some of the values are saved as
         # bytes directly, like pickle serialized values, which may raise an
         # exception when decoded with 'utf-8'.
-        self.client = redis.Redis(host=self.endpoint, port=self.port, db=self.db,
-                                  password=self.password, decode_responses=False,
-                                  socket_connect_timeout=self.create_connection_timeout,
-                                  max_connections=self.pool_max_size)
+        connection_pool = redis.ConnectionPool(
+            host=self.endpoint, port=self.port, db=self.db,
+            password=self.password, decode_responses=False,
+            socket_connect_timeout=self.create_connection_timeout,
+            max_connections=self.pool_max_size,
+            **connection_pool_kwargs
+        )
+        self.client = redis.Redis(connection_pool=connection_pool)
 
     async def _get(self, key, encoding="utf-8", _conn=None):
         value = await self.client.get(key)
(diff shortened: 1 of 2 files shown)


Dreamsorcerer:
The example here suggests we only need to pass `ssl=True`:
https://redis-py.readthedocs.io/en/stable/examples/ssl_connection_examples.html#Connecting-to-a-Redis-instance-via-SSL

I'm wondering if we should maybe just accept arbitrary arguments and pass them through to `Redis()` instead. Then the user can configure however they like.

cancan101:
The single SSL parameter is not enough in the general sense. There are a set of parameters specific to SSL, for example to ignore checking the certificate chain, which is required on heroku.

In addition there is a strong case to be made to allow specifying the connection pool class, for example, to allow turnib on blocking mode.

Dreamsorcerer:
More reason to just pass all arguments through, rather than trying to maintain a list of arguments.

Dreamsorcerer:
My first thought is to just copy all the existing arguments into a dict, which can be used as default values. Then just do `defaults.update(kwargs)` and pass that to `Redis()`.

That would disallow positional arguments, but not sure that's a major loss.

cancan101:
Right and I have connection_pool_kwargs. One of the parameters is specifically overridden per comment and a couple of the others are remapped based on names that this library uses more generally but sure there might be one or two that can be factored out as a general spread through

Dreamsorcerer:
I don't think there's any need for naming consistency, ultimately they are arguments for `Redis()`, not our cache (`MemoryCache` doesn't use any of the arguments).

Just realised we probably can't use `kwargs` though, because some arguments are needed for `BaseCache`. So, maybe just have a single `redis_args` parameter which takes a dict of arguments to be passed in.

cancan101:
Right, that is why I did the `connection_pool_kwargs`. That being said, I totally follow how you would suggest changing what I have in the PR?

Dreamsorcerer:
> Right, that is why I did the `connection_pool_kwargs`.

If we just allow arbitrary arguments for the entire `Redis()` class, then this awkwardness is unneeded. A user can simple do:
`redis_args={"connection_pool": SSLConnection(...)}`

Dreamsorcerer:
Maybe this PR is reasonable as a backwards-compatible approach for 0.x though.

If you're happy to go with that, then please retarget this PR to the 0.12 branch and create a new one for master that uses `redis_args`.

cancan101:
I want to make sure I understand what you are suggesting for the change to master. At present, a redis Cache would be instantiated as:
```python
cache = Cache(
  Cache.REDIS,
  endpoint="127.0.0.1",
  port=6379,
  password="xyz"
)
```
with this PR (as is), it could be instantiated as (for SSL + Blocking support):
```python
cache = Cache(
  Cache.REDIS,
  endpoint="127.0.0.1",
  port=6379,
  password="xyz",
  ssl=True,
  connection_pool_class=redis.BlockingConnectionPool,
  connection_pool_kwargs={"ssl_cert_reqs": None}
)
```
and it seems like you are suggesting:
```python
cache = Cache(
  Cache.REDIS,
  redis_args={"connection_pool": BlockingConnectionPool(dict(
    host="127.0.0.1",
    port=6379,
    password="xyz",
    decode_responses=False, # make sure to not forget this!
    connection_class=redis.SSLConnection
    ssl_cert_reqs=None
  ))}
)
```
which IMHO is not great DX and it makes the params very different for different Cache types.

Dreamsorcerer:
> it makes the params very different for different Cache types.

The params _are_ different for different cache types..
You can't just expect to pass the same connection params for a redis backend to a memcached backend (or some other backend) and have them work. This library literally just passes them to another library, so we have little control over how they work or what values are valid. I can't see a reason why they'd be expected to be the same.

There are `BaseCache` parameters which are the same across caches, and then there are connection parameters (maybe we can call it `conn_args` for consistency, rather than `redis_args`) which depend entirely on the backend library and could be vastly different from one cache to another.

By doing something like this, the advantages are that the user has full flexibility to pass any valid arguments desired to customise the backend, the code is kept very simple resulting in not needing to make changes to keep up-to-date with the upstream library and we don't get more bug reports when another user wants to change SSL ciphers or whatever other details come up in future.

cancan101:
>then there are connection parameters

Ideally there would be a way to still parse the DB url to get a dict rather than an instantiation of the Cache itself.
```python
cache_params = Cache.params_from_url(cache_url)
# For: rediss://:pass@host:port/db (injected by Heroku for example)
#-> {
    "host": "127.0.0.1",
    "port": 6379,
    "password": "xyz",
    "decode_responses": False, # okay, but where does this come from?
    "connection_class": redis.SSLConnection, # would be cool to get this from the url
    #ssl_cert_reqs=None # fine, this would have to be set by the user since is Heroku specific
}
```
Having a nice utility function like that would be nice and perhaps act a bridge between to the new API .

cancan101:
I do think there is value is splitting up the parsing of the URL from the instantiation. i.e. `Cache.from_url()` currently parses and instantiates which means if I want to inject Redis specific kwargs, that isn't possible. In the two step, I can optionally inject the Redis specific `kwargs`. This comes for example in that locally and in unit testing I use in memory cache (`memory://`) where as the deployed site uses redis. Since Heroku generates the URL, I can't just modify the redis URL to insert the blocking setting (which isn't easily usable anyways as I need to specify a class rather than a string).

There are still warts / leaks in the abstraction, for example the `decode_responses` needing to be set to `False` by default.

rgbkrk:
As an aside, now that I'm hitting this issue -- should `aiocache` be using `aioredis`, from the same org `aio-libs`?

Dreamsorcerer:
aioredis migrated to upstream redis. aioredis is no longer supported.

rgbkrk:
Rad, thank you!

Dreamsorcerer:
> How about passing a connection pool instance as an argument?

That's exactly what I suggested in my change request:
https://github.com/aio-libs/aiocache/pull/691#issuecomment-1470339883

It might make sense to just remove all these arguments and have the user pass in a `Redis` instance all the time. Would simplify what needs to be maintained here considerably, while providing the maximum amount of configuration. I don't think it adds much for the user either.

The only potential issue is that it appears we need `decode_responses=False` (which is the default). It looks like we could atleast assert it with `client.connection_pool.connection_kwargs.get("decode_responses")`.

Dreamsorcerer:
Again, if someone wants to create a new PR that accepts just accepts Redis client, I'll look at merging that into the master branch.
This PR can be retargeted to the 0.12 branch as a temporary fix that's backwards compatible.

antont:
> It might make sense to just remove all these arguments and have the user pass in a `Redis` instance all the time.

I did this for us now by subclassing:

```
import aiocache

class AiocacheRedisCacheForClient(aiocache.RedisCache):
    def __init__(self, client, **kwargs):
        """skips RedisCache.__init__"""
        BaseCache.__init__(self, **kwargs)
        self.client = client
```

I also encountered exactly the problem with `decode_responses=False`, because the pool I had created for fastapi-users has that option enabled, which is required there. This was a bit tricky to notice first. I tried using NullSerializer and other things, but finally resorted to creating two pools, one with responses decoded and the other not. It's still nice how we can have the config options in one place, where both pools are created.

Dreamsorcerer:
We've updated 1.0 to just accept a Redis client now.
@cancan101 If you'd like to retarget this PR to the 0.12 branch, I'm happy to merge this in if it's needed in the meantime.

cancan101:
Is there any way to re-open the PR?

cancan101:
ok, branch changed and rebased. lmk your thoughts.

Dreamsorcerer:
Looks good to me, want to mark the PR as ready?

Dreamsorcerer:
Looks like we have a few failing tests too, if you have some time to look at those?

cancan101:
Ok tests fixed

Issue #724, opened 2023-06-10
Make it possible to run tests with a subset of dependencies


mgorny:
Currently most of the tests seem to be unconditionally import one or more backends, effectively making it impossible to run the test suite without all of the optional dependencies installed. Would you accept a patch to make all of the backends optional in tests, and have pytest skip the respective tests when the dependencies of a specific backend are not available?

Dreamsorcerer:
If it's simple, sure. I think it might be possible to just modify this fixture in some way: https://github.com/aio-libs/aiocache/blob/master/tests/acceptance/conftest.py#L43

mgorny:
Thanks. It's not trivial but it isn't very complex either. I've made #725 for initial discussion of the approach.

mgorny:
Could you make a new release including my changes, please?

Dreamsorcerer:
Sorry for the delay, pushed now.

mgorny:
Thanks a lot! I can confirm that it works now for us in Gentoo.

PR #725, opened 2023-06-11
Make aiomcache and redis deps optional for unit tests [WIP]

mgorny:
## What do these changes do?

Make it possible to run a subset of unit tests without aiomcache and/or redis installed.  This makes testing easier for distributions that aren't planning to use these backends.

The changes are limited to the test suite.

## Related issue number

#724 

## Checklist

### DIFF
--- a/tests/ut/conftest.py
+++ b/tests/ut/conftest.py
@@ -5,8 +5,6 @@ from unittest.mock import create_autospec, patch
 import pytest
 
 from aiocache import caches
-from aiocache.backends.memcached import MemcachedCache
-from aiocache.backends.redis import RedisCache
 from aiocache.plugins import BasePlugin
 from ..utils import AbstractBaseCache, ConcreteBaseCache
 
@@ -61,11 +59,15 @@ def base_cache():
 
 @pytest.fixture
 async def redis_cache():
+    from aiocache.backends.redis import RedisCache
+
     async with RedisCache() as cache:
         yield cache
 
 
 @pytest.fixture
 async def memcached_cache():
+    from aiocache.backends.memcached import MemcachedCache
+
     async with MemcachedCache() as cache:
         yield cache
--- a/tests/ut/test_factory.py
+++ b/tests/ut/test_factory.py
@@ -29,10 +29,12 @@ else:
     CACHE_NAMES.append(Cache.REDIS.NAME)
 
 
+@pytest.mark.redis
 def test_class_from_string():
     assert _class_from_string("aiocache.RedisCache") == RedisCache
 
 
+@pytest.mark.redis
 def test_create_simple_cache():
     redis = _create_cache(RedisCache, endpoint="127.0.0.10", port=6378)
 
@@ -42,15 +44,15 @@ def test_create_simple_cache():
 
 
 def test_create_cache_with_everything():
-    redis = _create_cache(
-        RedisCache,
+    cache = _create_cache(
+        SimpleMemoryCache,
         serializer={"class": PickleSerializer, "encoding": "encoding"},
         plugins=[{"class": "aiocache.plugins.TimingPlugin"}],
     )
 
-    assert isinstance(redis.serializer, PickleSerializer)
-    assert redis.serializer.encoding == "encoding"
-    assert isinstance(redis.plugins[0], TimingPlugin)
+    assert isinstance(cache.serializer, PickleSerializer)
+    assert cache.serializer.encoding == "encoding"
+    assert isinstance(cache.plugins[0], TimingPlugin)
 
 
 class TestCache:
@@ -177,6 +179,7 @@ class TestCacheHandler:
     def test_create_not_reuse(self):
         assert caches.create("default") is not caches.create("default")
 
+    @pytest.mark.redis
     def test_create_extra_args(self):
         caches.set_config(
             {
@@ -193,6 +196,7 @@ class TestCacheHandler:
         assert cache.endpoint == "127.0.0.10"
         assert cache.db == 10
 
+    @pytest.mark.redis
     def test_retrieve_cache(self):
         caches.set_config(
             {
@@ -222,6 +226,7 @@ class TestCacheHandler:
         assert cache.serializer.encoding == "encoding"
         assert len(cache.plugins) == 2
 
+    @pytest.mark.redis
     def test_retrieve_cache_new_instance(self):
         caches.set_config(
             {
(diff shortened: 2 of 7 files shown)


mgorny:
This is my initial PoC for #724. I'd appreciate if you looked at it before I extend it to other kinds of tests.

Note that I've used `pytest.importorskip()` API as that's the usual way of skipping tests when some module is not available. If you prefer, I can switch to explicit `try:` / `except ImportError:` blocks instead. This may be actually necessary to fix PEP 8 errors because of `importorskip()` above other imports.

As a side effect from these changes, `tests/ut/test_factory.py` now verifies that `Cache.MEMCACHED` and `Cache.REDIS` are `None` when the relevant modules fail to import.

tests/ut/backends/test_redis.py
```diff
@@ -1,6 +1,9 @@
 from unittest.mock import ANY, AsyncMock, create_autospec, patch
 
 import pytest
+
+pytest.importorskip("redis")
```
Dreamsorcerer, line 5: I think this approach might be OK (maybe add a flake8 ignore to .flake8 for all of tests/ for the import ordering).

tests/ut/backends/test_memcached.py
```diff
@@ -1,8 +1,9 @@
 from unittest.mock import AsyncMock, patch
 
-import aiomcache
 import pytest
 
+aiomcache = pytest.importorskip("aiomcache")
```
Dreamsorcerer, line 5: But, this...

tests/ut/conftest.py
```diff
@@ -61,11 +59,13 @@ def base_cache():
...
+    redis = pytest.importorskip("aiocache.backends.redis")
+    async with redis.RedisCache() as cache:
         yield cache
 
 
 @pytest.fixture
 async def memcached_cache():
-    async with MemcachedCache() as cache:
+    memcached = pytest.importorskip("aiocache.backends.memcached")
+    async with memcached.MemcachedCache() as cache:
```
Dreamsorcerer, line 70: ....and this, are more problematic.

`pytest.importorskip()` returns `Any`, which loses all our type checking features when we check with mypy. We use this to help verify that users won't encounter type errors when using the library correctly (as tests/examples are the only place we actually test/use the public API).

Dreamsorcerer (commented):
My only other concern is that this doesn't seem to have an explicit flag or anything?
It'd probably be nice to be explicit with which tests are expected to run (e.g. `--all` or `--redis` or something, so it errors if the requested tests would get skipped).

Maybe not a big issue, as codecov reports should highlight if something goes wrong, but would be nice to have.

tests/ut/conftest.py
mgorny: Ah, this never occurred to me! I suppose we could either do something like:

    pytest.importorskip(...)  # just to skip test if it's missing
    from ... import ...  # the actual import

or:

    try:
        from ... import ...
    except ImportError:
        pytest.skip("x is missing")

Which one would you prefer?

mgorny:
> It'd probably be nice to be explicit with which tests are expected to run (e.g. --all or --redis or something, so it errors if the requested tests would get skipped).

Hmm, I think the customary way of doing this would be to use pytest marks, i.e. mark tests as "memcached tests", "redis tests", etc. appropriately. Then I could try adding some special options, or alternatively just run all tests by default and support opting out of specific tests e.g. via `-m "not redis"`. The latter will probably be easier.

I suppose the mark approach also removes the type problem mentioned above. I think we can also apply marks from conftest to specific files, so maybe it would also resolve the problem of import order.

tests/ut/conftest.py
Dreamsorcerer: I think the first is preferable.

mgorny:
Ok, I've pushed another approach entirely. For the most part, this one defaults to running all unittests but adds `memcached` and `redis` markers to skip specific tests, and moves imports so that they could be skipped without causing `ImportError`s. Now we could run e.g.:

    python -m pytest tests/ut -m "not memcached and not redis"

The fourth commit is perhaps the most scary since the relevant thingies need to be imported in every test function separately. Technically, it's optional — if that's too much, I can remove it. We'd then have to explicitly pass e.g. `--ignore tests/ut/backend/test_memcached.py` but that's definitely doable for us.

tests/ut/test_factory.py
```diff
@@ -30,7 +45,7 @@ def test_create_simple_cache():
 
 def test_create_cache_with_everything():
     redis = _create_cache(
-        RedisCache,
+        SimpleMemoryCache,
```
mgorny, line 48: I've replaced `RedisCache` with `SimpleMemoryCache` here since it seemed not to feature anything redis-specific.

Dreamsorcerer: OK, variable should be renamed though.

Dreamsorcerer:
> The fourth commit is perhaps the most scary since the relevant thingies need to be imported in every test function separately. Technically, it's optional — if that's too much, I can remove it. We'd then have to explicitly pass e.g. `--ignore tests/ut/backend/test_memcached.py` but that's definitely doable for us.

Yeah, can't say I'm a fan of the copy/paste of imports in every test.

tests/ut/test_factory.py
mgorny: Right, fixed.

mgorny:
> Yeah, can't say I'm a fan of the copy/paste of imports in every test.

Should I remove that commit then?

Dreamsorcerer:
Yeah, if it still works for you, I'd rather avoid people needing to add imports to every each new test.

mgorny:
Done. Please let me know whether you have any further remarks before I proceed to see if I can get other tests handled similarly.

Dreamsorcerer:
Looks good to me.

mgorny:
Thanks. I've updated the PR to cover acceptance tests as well. I've skipped performance tests because they would be harder than that and we aren't going to run them anyway.

I've also confirmed that this works in all 4 cases: no redis/memcached, memcached only, redis only and memcached+redis.

tests/ut/test_factory.py
```diff
@@ -3,23 +3,38 @@
...
-assert Cache.REDIS is not None
-assert Cache.MEMCACHED is not None
-CACHE_NAMES = (Cache.MEMORY.NAME, Cache.REDIS.NAME, Cache.MEMCACHED.NAME)
 
+CACHE_NAMES = [Cache.MEMORY.NAME]
 
+try:
+    from aiocache.backends.memcached import MemcachedCache
+except ImportError:
+    MemcachedCache = None
```
Dreamsorcerer, line 18: It would probably be nice to have a test run that does cover this. Can figure it out later though.

mgorny: I'll try to update the test matrix in a minute.

mgorny:
Thanks for merging!

PR #731, opened 2023-06-21
Add workflows with partial backends to CI

mgorny:
## What do these changes do?

Attempt to add two new CI workflows, one testing without memcached and the other without redis.

No.

## Related issue number

Suggested in #725

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -99,11 +99,18 @@ jobs:
     - name: Run unittests
       env:
         COLOR: 'yes'
-      run: |
-        pytest tests --cov-report xml --cov-report html
-        python -m coverage xml
+      run: pytest tests --cov-report xml --cov-report html
     - name: Run functional tests
       run: bash examples/run_all.sh
+    - name: Uninstall optional backends
+      run: pip uninstall -y aiomcache redis
+    - name: Run unittests with minimal backend set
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -111,6 +111,54 @@ jobs:
         flags: unit
         fail_ci_if_error: false
 
+  test-partial-backends:
```
Dreamsorcerer, line 114: I wonder if it'd be simpler to just add an extra step to the existing test job, which will probably help to keep all the coverage steps together etc.

Dreamsorcerer: I don't think a matrix will add much. We can just uninstall both libraries and do 1 extra test run.

mgorny:
Turned out to be surprisingly easy. I went for removing them after installing to avoid changing `requirements.txt` since we still want them installed by default, and they require special pytest arguments anyway.

.github/workflows/ci.yml
mgorny: Sure, will update.

mgorny:
Done. FWICS the 'else' clauses are now green in codecov, so I suppose it worked ;-).

Dreamsorcerer:
Awesome, thanks.

Issue #677, opened 2023-02-27
Remove factory


Dreamsorcerer:
factoyr.py can probably be removed. It's overly complicated and causes typing issues while seeming to provide no benefit at all.

I think it's probably preferable to simply import the caches the user wants explicitly:
```
from aiocache.backends.redis import RedisCache
```

mikeoconnor0308:
Do you want to remove the support for CacheHandler / aliases as well?

Dreamsorcerer:
Yes, I don't think the config adds much use and makes it impossible to fix #609.

PR #755, opened 2023-08-18
Change RedisBackend to accept Redis client directly

### LINKED ISSUE #550: SSL/TLS support for redis cache
The `aioredis` library supports connecting to redis using TLS, and has done since version 0.2.6 (see https://github.com/aio-libs/aioredis-py/blob/master/CHANGELOG.md#026-2016-03-30). 

The `aiocache` implementation of the redis backend, however, does not support TLS - it doesn't use the `ssl` parameter when initialising the pool here: https://github.com/aio-libs/aiocache/blob/master/aiocache/backends/redis.py#L209 

You can see the function signature of create_pool here: https://github.com/moserware/aioredis/blob/master/aioredis/pool.py#L14

```python
@asyncio.coroutine
def create_pool(address, *, db=0, password=None, ssl=None, encoding=None,
                minsize=10, maxsize=10, commands_factory=Redis, loop=None):
    ...
```

Is there a reason `aiocache` does not support TLS as a config option for a redis backend? If not, please could support be added?

ben-cutler-datarobot:
## What do these changes do?
These changes implement the feedback from this PR:
https://github.com/aio-libs/aiocache/pull/691#issuecomment-1587624406
Fixes #550 

This PR removes the construction of the redis.Redis object from the RedisBackend class and makes it a dependency. This allows users of the library to construct their Redis instances which might have to have a custom SSL cert, or be a more complex use case.  This allows apps using this package to only have one redis instance to manage if they want to use redis for other things.

This PR also does a few other things (unfortunately)
0. It changes the constructor of the RedisBackend to take in a client instead of client args. This means that the `RedisBackend` class's public members like `port` will be moved to deeper in the code. 
1. it renames `endpoint` to `host`. This naming is consistent with [redis](https://github.com/redis/redis-py/blob/28cc65c18cc4fb37ef14497c963eb181dba8d25d/redis/asyncio/client.py#L173) and your [aiomcache library](https://github.com/aio-libs/aiomcache/blob/083ea0627ef17245aa6ff165a224b5a923f631e1/aiomcache/client.py#L61)
2. It renames the redis URLstrings `pool_max_size` --> `pool_max_size` and `create_connection_timeout` --> `create_connection_timeout` to be consistent with the naming in redis-py. 
3. It removes the `pool_min_size` from the RedisBackend constructor (I assume I may as well remove it) 

## Related issue number

Fixes #691

## Checklist

### DIFF
--- a/aiocache/backends/redis.py
+++ b/aiocache/backends/redis.py
@@ -1,5 +1,4 @@
 import itertools
-import warnings
 from typing import Any, Callable, Optional, TYPE_CHECKING
 
 import redis.asyncio as redis
@@ -38,41 +37,11 @@ class RedisBackend(BaseCache[str]):
 
     def __init__(
         self,
-        endpoint="127.0.0.1",
-        port=6379,
-        db=0,
-        password=None,
-        pool_min_size=_NOT_SET,
-        pool_max_size=None,
-        create_connection_timeout=None,
+        client,
         **kwargs,
     ):
         super().__init__(**kwargs)
-        if pool_min_size is not _NOT_SET:
-            warnings.warn(
-                "Parameter 'pool_min_size' is deprecated since aiocache 0.12",
-                DeprecationWarning, stacklevel=2
-            )
-
-        self.endpoint = endpoint
-        self.port = int(port)
-        self.db = int(db)
-        self.password = password
-        # TODO: Remove int() call some time after adding type annotations.
-        self.pool_max_size = None if pool_max_size is None else int(pool_max_size)
-        self.create_connection_timeout = (
-            float(create_connection_timeout) if create_connection_timeout else None
-        )
-
-        # NOTE: decoding can't be controlled on API level after switching to
-        # redis, we need to disable decoding on global/connection level
-        # (decode_responses=False), because some of the values are saved as
-        # bytes directly, like pickle serialized values, which may raise an
-        # exception when decoded with 'utf-8'.
-        self.client = redis.Redis(host=self.endpoint, port=self.port, db=self.db,
-                                  password=self.password, decode_responses=False,
-                                  socket_connect_timeout=self.create_connection_timeout,
-                                  max_connections=self.pool_max_size)
+        self.client = client
 
     async def _get(self, key, encoding="utf-8", _conn=None):
         value = await self.client.get(key)
@@ -175,9 +144,6 @@ class RedisBackend(BaseCache[str]):
     async def _redlock_release(self, key, value):
         return await self._raw("eval", self.RELEASE_SCRIPT, 1, key, value)
 
-    async def _close(self, *args, _conn=None, **kwargs):
-        await self.client.close()
-
     def build_key(self, key: str, namespace: Optional[str] = None) -> str:
         return self._str_build_key(key, namespace)
 
@@ -196,24 +162,21 @@ class RedisCache(RedisBackend):
         the backend. Default is an empty string, "".
     :param timeout: int or float in seconds specifying maximum timeout for the operations to last.
         By default its 5.
-    :param endpoint: str with the endpoint to connect to. Default is "127.0.0.1".
-    :param port: int with the port to connect to. Default is 6379.
-    :param db: int indicating database to use. Default is 0.
-    :param password: str indicating password to use. Default is None.
-    :param pool_max_size: int maximum pool size for the redis connections pool. Default is None.
-    :param create_connection_timeout: int timeout for the creation of connection. Default is None
+    :param client: redis.Redis which is an active client for working with redis
     """
 
     NAME = "redis"
 
     def __init__(
         self,
+        client: redis.Redis,
         serializer: Optional["BaseSerializer"] = None,
         namespace: str = "",
         key_builder: Callable[[str, str], str] = lambda k, ns: f"{ns}:{k}" if ns else k,
         **kwargs: Any,
     ):
         super().__init__(
+            client=client,
             serializer=serializer or JsonSerializer(),
             namespace=namespace,
             key_builder=key_builder,
@@ -237,4 +200,5 @@ class RedisCache(RedisBackend):
         return options
 
     def __repr__(self):  # pragma: no cover
-        return "RedisCache ({}:{})".format(self.endpoint, self.port)
+        connection_kwargs = self.client.connection_pool.connection_kwargs
+        return "RedisCache ({}:{})".format(connection_kwargs['host'], connection_kwargs['port'])
--- a/aiocache/backends/redis.py
+++ b/aiocache/backends/redis.py
@@ -37,7 +37,7 @@ class RedisBackend(BaseCache[str]):
 
     def __init__(
         self,
-        client,
+        client: redis.Redis,
         **kwargs,
     ):
         super().__init__(**kwargs)
--- a/aiocache/backends/redis.py
+++ b/aiocache/backends/redis.py
@@ -41,6 +41,14 @@ class RedisBackend(BaseCache[str]):
         **kwargs,
     ):
         super().__init__(**kwargs)
+
+        # NOTE: decoding can't be controlled on API level after switching to
+        # redis, we need to disable decoding on global/connection level
+        # (decode_responses=False), because some of the values are saved as
+        # bytes directly, like pickle serialized values, which may raise an
+        # exception when decoded with 'utf-8'.
+        if client.connection_pool.connection_kwargs['decode_responses']:
+            raise ValueError("redis client must be constructed with decode_responses set to False")
         self.client = client
 
     async def _get(self, key, encoding="utf-8", _conn=None):
--- a/aiocache/factory.py
+++ b/aiocache/factory.py
@@ -3,7 +3,7 @@ import urllib
 from copy import deepcopy
 from typing import Dict
 
-import redis
+import redis.asyncio as redis
 
 from aiocache import AIOCACHE_CACHES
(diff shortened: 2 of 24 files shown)


aiocache/backends/redis.py
```diff
@@ -38,41 +37,11 @@ class RedisBackend(BaseCache[str]):
...
-        create_connection_timeout=None,
+        client,
         **kwargs,
     ):
         super().__init__(**kwargs)
-        if pool_min_size is not _NOT_SET:
-            warnings.warn(
-                "Parameter 'pool_min_size' is deprecated since aiocache 0.12",
-                DeprecationWarning, stacklevel=2
-            )
```
ben-cutler-datarobot, line 55: I assume if we are changing the function signature we may as well get rid of this

aiocache/backends/redis.py
```diff
@@ -175,9 +144,6 @@ async def _raw(self, command, *args, encoding="utf-8", _conn=None, **kwargs):
     async def _redlock_release(self, key, value):
         return await self._raw("eval", self.RELEASE_SCRIPT, 1, key, value)
 
-    async def _close(self, *args, _conn=None, **kwargs):
-        await self.client.close()
-
```
ben-cutler-datarobot, line 180: Since the construction of the redis.Redis is no longer done in this class, then this class shouldn't have a method to close it.

aiocache/factory.py
```diff
@@ -112,15 +121,21 @@ def from_url(cls, url):
...
             kwargs["password"] = parsed_url.password
 
-        return Cache(cache_class, **kwargs)
+        for arg in ['max_connections', 'socket_connect_timeout']:
+            if arg in kwargs:
+                kwargs[arg] = int(kwargs[arg])
+        if cache_class == cls.REDIS:
+            return Cache(cache_class, client=redis.Redis(**kwargs))
+        else:
+            return Cache(cache_class, **kwargs)
```
ben-cutler-datarobot, line 138: ditto on the clunkyness around the looking at the cache class.

tests/ut/backends/test_redis.py
```diff
@@ -29,64 +29,6 @@ def redis():
...
-            "port": 6379,
-            "max_connections": 10,
-            "socket_connect_timeout": 1.5,
-        })
-        mock_class.assert_called_with(**kwargs)
-
-        assert redis_backend.db == 2
-        assert redis_backend.port == 6379
-        assert redis_backend.pool_max_size == 10
-        assert redis_backend.create_connection_timeout == 1.5
```
ben-cutler-datarobot, line 89: I got rid of these tests because they don't really test anything if the client construction responsibility is moved out of the RedisBackend class.

aiocache/backends/redis.py
```diff
@@ -38,41 +37,11 @@ class RedisBackend(BaseCache[str]):
...
-        self.pool_max_size = None if pool_max_size is None else int(pool_max_size)
-        self.create_connection_timeout = (
-            float(create_connection_timeout) if create_connection_timeout else None
-        )
-
-        # NOTE: decoding can't be controlled on API level after switching to
-        # redis, we need to disable decoding on global/connection level
-        # (decode_responses=False), because some of the values are saved as
-        # bytes directly, like pickle serialized values, which may raise an
-        # exception when decoded with 'utf-8'.
```
Dreamsorcerer, line 71: I think this is important. Can we inspect the client and raise an exception if it is not configured correctly?

aiocache/factory.py
```diff
@@ -29,10 +31,17 @@ def _create_cache(cache, serializer=None, plugins=None, **kwargs):
...
+    if cache == AIOCACHE_CACHES.get("redis"):
+        return cache(
+            serializer=serializer,
+            plugins=plugins_instances,
+            namespace=kwargs.pop('namespace', ''),
+            ttl=kwargs.pop('ttl', None),
+            client=redis.Redis(**kwargs)
+        )
+    else:
+        return cache(serializer=serializer, plugins=plugins_instances, **kwargs)
```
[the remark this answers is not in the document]
Dreamsorcerer: I don't think this has any utility and can be removed before the 1.0 release (See #677).

tests/conftest.py
```diff
@@ -0,0 +1,22 @@
+import pytest
+import redis.asyncio as redis
+
+
+@pytest.fixture()
+def max_conns():
```
Dreamsorcerer, line 6: Does this serve a purpose?

tests/conftest.py
```diff
@@ -0,0 +1,22 @@
...
+        host="127.0.0.1",
+        port=6379,
+        db=0,
+        password=None,
+        decode_responses=False,
+        socket_connect_timeout=None,
+        max_connections=max_conns
+    )
+    yield r
+    await r.close()
```
Dreamsorcerer, line 22:
```suggestion
    aync with redis.Redis(
        host="127.0.0.1",
        port=6379,
        db=0,
        password=None,
        decode_responses=False,
        socket_connect_timeout=None,
        max_connections=max_conns
    ) as r:
        yield r
```

Dreamsorcerer:
Looks good to me.

aiocache/backends/redis.py
ben-cutler-datarobot: :+1: Added this check:

https://github.com/aio-libs/aiocache/pull/755/commits/b6474be69e32effbda4c585c4b8f3d77e81f2f23#diff-d343e8d4d6f140b9992b916ca117cdb4549de4e0bd36628cc94f2b3627ba2b88R44-R51

and tested here:

https://github.com/aio-libs/aiocache/pull/755/commits/b6474be69e32effbda4c585c4b8f3d77e81f2f23#diff-ed4918b243ad91403f48326ac0e674ad453c94754001c35e3c79334df40ac509R33-R41

tests/conftest.py
ben-cutler-datarobot: Yeah in `performance/conftest.py` the Redis there was constructed with `pool_max_size` set to 1, so I wanted to keep that implementation the same. I think it made the most sense to have only one redis fixture, so I just made the main fixture take it as an argument that can be modified.

tests/conftest.py
ben-cutler-datarobot: That's a better approach, thanks

Dreamsorcerer:
Seems there still some issues with the examples and the factory code. I'm wondering if it'll be easier to just create a PR for #677 first, and then update this one.

examples/conftest.py
Dreamsorcerer, line 1: These aren't pytest tests, they seem to be examples that are executed in the test run to ensure they are not completely broken. So, I don't think it makes sense to have a conftest.py file, and we should try to keep each example self-contained.

Also, aren't all these values the default values in Redis? So, we should be able to just drop them entirely and still have it work..

ben-cutler-datarobot: Yeah, it looks like the extra kwargs to the client for talking to redis in docker-compose aren't needed 😅.

tests/conftest.py
```diff
@@ -0,0 +1,26 @@
+import pytest
+import redis.asyncio as redis
```
Dreamsorcerer, line 2: Ah, nearly through the tests. Seems there's another test run after uninstalling redis, so this might need tweaking to get that working.

ben-cutler-datarobot: Let me try moving the import into the fixture.

examples/cached_alias_config.py
```diff
@@ -45,17 +47,17 @@ async def alt_cache():
     assert isinstance(cache, Cache.REDIS)
     assert isinstance(cache.serializer, PickleSerializer)
     assert len(cache.plugins) == 2
-    assert cache.endpoint == "127.0.0.1"
-    assert cache.timeout == 1
-    assert cache.port == 6379
+    assert cache.client.connection_pool.connection_kwargs["host"] == "127.0.0.1"
+    assert cache.client.connection_pool.connection_kwargs["socket_connect_timeout"] == 1
+    assert cache.client.connection_pool.connection_kwargs["port"] == 6379
```
Dreamsorcerer, line 52:
```suggestion
    connection_args = cache.client.connection_pool.connection_kwargs
    assert connection_args["host"] == "127.0.0.1"
    assert connection_args["socket_connect_timeout"] == 1
    assert connection_args["port"] == 6379
```

tests/acceptance/test_factory.py
```diff
@@ -11,22 +11,22 @@ async def test_from_url_memory(self):
...
-            assert cache.endpoint == "endpoint"
-            assert cache.port == 1000
-            assert cache.password == "pass"
-            assert cache.pool_max_size == 50
-            assert cache.create_connection_timeout == 20
+            assert cache.client.connection_pool.connection_kwargs["host"] == "endpoint"
+            assert cache.client.connection_pool.connection_kwargs["port"] == 1000
+            assert cache.client.connection_pool.connection_kwargs["password"] == "pass"
+            assert cache.client.connection_pool.max_connections == 50
+            assert cache.client.connection_pool.connection_kwargs["socket_connect_timeout"] == 20
```
Dreamsorcerer, line 29:
```suggestion
            connection_args = cache.client.connection_pool.connection_kwargs
            assert connection_args["host"] == "endpoint"
            assert connection_args["port"] == 1000
            assert connection_args["password"] == "pass"
            assert cache.client.connection_pool.max_connections == 50
            assert connection_args["socket_connect_timeout"] == 20
```

Dreamsorcerer:
Great, thanks for getting all of that working!

PR #807, opened 2024-01-21
fix: prevent memory leak in _redlock_release of SimpleMemoryBackend

moshego189:
## What do these changes do?

The changes refactor the `_redlock_release` method in the `SimpleMemoryBackend` class. Instead of directly removing items from the `_cache` dictionary using the `pop` method, the method now invokes the `__delete` method. This ensures that when an item is released from the cache, associated timer handles in the `_handlers` dictionary are also properly cancelled, thereby preventing potential memory leaks.

There are no direct changes in behavior for the end user. This update enhances the internal memory management of the cache system, contributing to improved stability and efficiency without altering the functional interface.

## Checklist

### DIFF
--- a/aiocache/backends/memory.py
+++ b/aiocache/backends/memory.py
@@ -94,8 +94,7 @@ class SimpleMemoryBackend(BaseCache[str]):
 
     async def _redlock_release(self, key, value):
         if self._cache.get(key) == value:
-            self._cache.pop(key)
-            return 1
+            return self.__delete(key)
         return 0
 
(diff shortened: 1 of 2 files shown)


Dreamsorcerer:
Could we have a test that validates the handler is no longer present after release (and ideally that the callback has been cancelled)?

moshego189:
> Could we have a test that validates the handler is no longer present after release (and ideally that the callback has been cancelled)?

Done!

Issue #834, opened 2024-04-01
UnicodeDecodeError when using MsgPackSerializer


nottmtt:
Hi!

When using MsgPack as the (de)serializer for aiocache (no matter the backend), the following error occurs when reading an item from the cache:

```
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x82 in position 0: invalid start byte
```

(I'm using Python 3.12.2, aiocache 0.12.2, aiomcache 0.8.1, redis 4.6.0 and msgpack 1.0.8.)

The error seems to stem from the fact that most backends (prematurely?) decode values from the cache before passing them to the (de)serializer (see [backends/memcached.py:24](https://github.com/aio-libs/aiocache/blob/master/aiocache/backends/memcached.py#L24)). Replacing this line with `return value` makes it work (for MsgPack, at least).

But I guess simply removing the value decoding logic isn't a solution. If anyone can give me a clue of what's going wrong here (besides the `value.decode()`), I'll gladly create a merge request with (my attempt at) a fix.

Minimal test case (using `MemcachedCache`, use `RedisCache` if you wish to):

```py
import asyncio

from aiocache               import MemcachedCache
from aiocache.serializers   import MsgPackSerializer

async def test():
    cache = MemcachedCache(
        serializer = MsgPackSerializer(),
    )

    await cache.set("test_case", { "key1": "test", "key2": 3 })

    test_case = await cache.get("test_case") # UnicodeDecodeError here
    print(test_case)

if __name__ == "__main__":
    asyncio.run(test())
```

Full error:

```
Traceback (most recent call last):
  File "/home/spike/aiocache_test.py", line 17, in <module>
    asyncio.run(test())
  File "/usr/lib/python3.12/asyncio/runners.py", line 194, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/asyncio/base_events.py", line 685, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/home/spike/aiocache_test.py", line 13, in test
    test_case = await cache.get("test_case")
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #836, opened 2024-04-01
Add MsgPackSerializer-related acceptance tests

nottmtt:
This pull request adds acceptance tests for `MsgPackSerializer` which were lacking from the current test-suite.

They _will_ break CI since there's currently issues when using `MsgPackSerializer` as the serializer, which these tests are here to catch.

See issue #834.

### DIFF
--- a/tests/acceptance/test_serializers.py
+++ b/tests/acceptance/test_serializers.py
@@ -2,6 +2,7 @@ import pickle
 import random
 from typing import Any
 
+import msgpack
 import pytest
 from marshmallow import Schema, fields, post_load
 
@@ -13,6 +14,7 @@ except ImportError:
 from aiocache.serializers import (
     BaseSerializer,
     JsonSerializer,
+    MsgPackSerializer,
     NullSerializer,
     PickleSerializer,
     StringSerializer,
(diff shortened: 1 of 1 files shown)


tests/acceptance/test_serializers.py
```diff
@@ -141,6 +143,28 @@ async def test_multi_set_multi_get_types(self, cache, obj):
...
 
 
+class TestMsgPackSerializer:
+    TYPES = (1, 2.0, "hi", True, ["1", 1], {"key": "value"})
+
+    @pytest.mark.parametrize("obj", TYPES)
+    async def test_set_get_types(self, cache, obj):
+        cache.serializer = MsgPackSerializer()
+        assert await cache.set(Keys.KEY, obj) is True
+        assert await cache.get(Keys.KEY) == msgpack.loads(msgpack.dumps(obj))
```
Dreamsorcerer, line 153: I think some of the failures are related to the test itself (specifically `msgpack.loads(msgpack.dumps(obj))`)?
https://github.com/aio-libs/aiocache/actions/runs/8513779973/job/23320587633?pr=836#step:7:82

nottmtt: I don't think the test itself is causing issues here; adding

```py
assert msgpack.loads(msgpack.dumps(obj)) == obj
```

right before the other asserts in `TestMsgPackSerializer` doesn't raise assertion errors.

(I just noticed that some tests fail with a `TypeError: a bytes-like object is required, not 'str'` error though, I wonder what's causing that.)

Dreamsorcerer: > My guess would be that `TypeError: a bytes-like object is required, not 'str'` is raised because of the same encoding issues as the `UnicodeDecodeError` ones. But it's not raised by msgpack itself, that's for sure.

That's the one I was linking to. ;)

Dreamsorcerer:
#1079

Issue #791, opened 2023-12-19
Preserve function signature with decorator


Wyko:
I've noticed that my IDE (VS Code) does not show the function arguments of anything decorated by the `@cached` decorator. It would be great if the decorator would fully preserve the function's signature.

![image](https://github.com/aio-libs/aiocache/assets/467094/098b9c0e-da43-4130-8a05-ff3c8929f9ba)

Dreamsorcerer:
#512

Wyko:
Does #512 really match the question? Talking about typing vs preserving the original function signature.

Dreamsorcerer:
You're showing type annotations from your IDE. We don't have complete type annotations (or any exposed at all), so your IDE won't give you the correct types.

Issue #512, opened 2020-10-08
Typing


JeppeKlitgaard:
Since `aiocache` only supports modern Python 3 versions, I was wondering if there would be an interesting in having the library be typed?

I wouldn't mind doing the work to get this done, but wanted to get feedback on whether this is something that the maintainers of `aiocache` would even want.

On that same note, how do people feel about dropping Python 3.5 support now that the release has officially reached end-of-life.

argaen:
Hey @JeppeKlitgaard, I don't have anything against adding typing so go for it 👍. Same with dropping 3.5 support, been a while since wanted to do this plus adding support for 3.8 (and 3.9 now) but if you want to give it a try, please do :)

Dreamsorcerer:
Mypy has been configured on the repository, so if anyone wants to start contributing to this, please feel free to make a PR that adds some more annotations.

To work on it, start by uncommenting `check_untyped_defs` in .mypy.ini and fixing the errors that appear when running `mypy`.
If there are too many errors to start with, you can disable the option in sections of the code initially until it is manageable. e.g.
```
[mypy-tests.*]
check_untyped_defs = False
```

Olegt0rr:
@Dreamsorcerer is there an option to split this issue?

1) add paramspec to the decorator 
2) add other typing and mypy checks

Looks like 1 has no blocker and can be published fast

Just added some strings for example and got expected behaviour:
<img width="324" alt="Screenshot 2024-09-02 at 23 47 13" src="https://github.com/user-attachments/assets/221aa5d8-de05-40f2-9dbe-6571a3851dd8">

Dreamsorcerer:
No, exposing type hints at all causes type errors in everyone's code with v0. I tried enabling it before, but had to disable it again.

The v1 milestone is listed in the order that tasks should be completed:
https://github.com/aio-libs/aiocache/milestone/8

Once mypy is running, we could potentially start taking typing contributions, though the other tasks listed before this one will ensure the easiest path to typing.

Olegt0rr:
Even single TypeVar? 
It looks fully abstract and compatible with any decorated functions

Just tried to run mypy with my example and it passed well

P.S.: I looked at v1 - there is a fairly large amount of improvements. I definitely won't be able to switch on quickly. At the same time, I would like to make the current version at least a little more convenient

Dreamsorcerer:
> Even single TypeVar?

[...]

PR #838, opened 2024-04-03
WIP: Adds typing to the decorators

fabiob:
## What do these changes do?

They add strong typing to `aiocache` decorator syntax.

Somewhat, as their IDEs will start to hint at the expected usage of the decorators.

## Related issue number

It is related to #512 and #791

## Checklist

- [x] I think the code is well written (still a WIP though)
- [ ] Unit tests for the changes exist (would love to get some guidance on that)
- [ ] Documentation reflects the changes (I'm not sure it is needed)


Dreamsorcerer:
Any typing improvements will need to wait until other tasks on the v1 milestone are complete. At the very least, we need mypy running, but that's currently blocked by 3 tasks preceding it: https://github.com/aio-libs/aiocache/milestone/8
(I ordered the tasks in the milestone so, for the most part, they need to be completed in that order to avoid issues).

guybortnikov:
@fabiob @Dreamsorcerer any update on this PR? It will help for any project that uses type checking

Dreamsorcerer:
See the previous comment: https://github.com/aio-libs/aiocache/pull/838#issuecomment-2037683694

PR #880, opened 2024-07-31
ADDED: typing to cached.__call__

GrafLearnt:
## What do these changes do?
Adds typing support for functions decorated with `@cached()`

No

## Related issue number
null

## Checklist

### DIFF
--- a/aiocache/decorators.py
+++ b/aiocache/decorators.py
@@ -2,12 +2,16 @@ import asyncio
 import functools
 import inspect
 import logging
+from typing import Any, Awaitable, Callable, ParamSpec, TypeVar
 
 from aiocache.base import SENTINEL
 from aiocache.factory import Cache, caches
 from aiocache.lock import RedLock
 
 logger = logging.getLogger(__name__)
+
+P = ParamSpec("P")
+T = TypeVar("T", bound=Any)
 
 
 class cached:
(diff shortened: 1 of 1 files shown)


Dreamsorcerer:
As mentioned in other issues/PRs, typing won't work in this library until the preceding steps in the milestone are complete: https://github.com/aio-libs/aiocache/milestone/8

Feel free to contribute by picking up the tasks starting from the top of that milestone.

aiocache/decorators.py
```diff
@@ -2,13 +2,17 @@
...
+from typing import Any, Awaitable, Callable, ParamSpec, TypeVar
 
 from aiocache.base import SENTINEL
 from aiocache.factory import Cache, caches
 from aiocache.lock import RedLock
 
 logger = logging.getLogger(__name__)
 
+P = ParamSpec("P")
+T = TypeVar("T", bound=Any)
```
Kilo59, line 14: You'll need to use `typing_extensions` to support older versions of python.

aiocache/decorators.py
```diff
@@ -2,13 +2,17 @@
 import functools
 import inspect
 import logging
+from typing import Any, Awaitable, Callable, ParamSpec, TypeVar
```
Kilo59, line 5: You'll need to use `typing_extensions` to support older versions of python.

PR #895, opened 2024-09-02
Add minimal typing for decorators

Olegt0rr:
## What do these changes do?

The task of adding full typing to the library is quite a long one. We can wait long enough for contributors to have time.

Since we are already using the library today, I suggest covering the smallest, but most frequently used part of it in the simplest way.

Even if it is more correct to describe ParamSpec, etc. - let's do it in this minimal version. And as soon as the contributors have the strength and time, we will redo it in the target form.

Displaying original type hints of decorated function

### Before

Screenshot 2024-09-03 at 00 03 47

### After

Screenshot 2024-09-03 at 00 04 26

## Related issue number

Related to #512 (as a temporary solution)

## Checklist

### DIFF
--- a/aiocache/decorators.py
+++ b/aiocache/decorators.py
@@ -2,12 +2,15 @@ import asyncio
 import functools
 import inspect
 import logging
+from typing import TypeVar
 
 from aiocache.base import SENTINEL
 from aiocache.factory import Cache, caches
 from aiocache.lock import RedLock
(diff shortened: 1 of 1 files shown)


Olegt0rr:
Closed, cause: https://github.com/aio-libs/aiocache/issues/512#issuecomment-2325312283

PR #927, opened 2024-11-22
feat: add invalidate_cache for @cached decorator

biraj21:
## What do these changes do?

This PR adds an `invalidate_cache` function to the `aiocache` library, allowing users to explicitly invalidate the cache of a function decorated with `@cache`. Users can call the method directly on the cached function object with the arguments that were used to generate the cache key. Usage:
```python
await <function_name>.invalidate_cache(*args, **kwargs)
```

Yes, this new functionality provides users with a convenient way to manually invalidate the cache for specific arguments of a cached function, enhancing flexibility and control over cache management.

## Related issue number

N/A

## Checklist

### DIFF
--- a/aiocache/decorators.py
+++ b/aiocache/decorators.py
@@ -15,6 +15,10 @@ class cached:
     Caches the functions return value into a key generated with module_name, function_name
     and args. The cache is available in the function object as ``<function_name>.cache``.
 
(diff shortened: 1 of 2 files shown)


Dreamsorcerer:
I'd prefer to avoid adding more hacks like this onto the old code. This should really wait for #625 to be completed, which may require the other tasks listed before it in the milestone: https://github.com/aio-libs/aiocache/milestone/8

PR #975, opened 2025-04-01
implement valkey support

### LINKED ISSUE #882: Redis to valkey?
I was wondering if there was any interest or ongoing work in migrating from redis to valkey? Or in adding valkey as a separate backend for this project?

amirreza8002:
## What do these changes do?
add support for valkey, using valkey-gilde

## Related issue number
Fixes #882 

## Checklist

- [x]  a usable valkey client
- [x] proper testing


amirreza8002:
notes about this PR:

1. valkey glide does not have a connection pool, so the implementation is a bit different from redis

2. i've left some small things in there for future (e.g: set returns "OK" instead of True)

3. some tests fail, at least in my local machine, i am yet to discover what the problem actually is, they seem to be flaky and hard to debug

4. i'm thinking of removing the lua scripts, they can be written in python (i did kinda remove one cause it was problematic)

5. i think we should add some sort of shortcut to create a valkey client, since glide's way is ugly

i'll continue to debug the problems with the tests, they could be something obvious that my tired eyes can't see or something hard, idk, i really appreciate any help and ideas and tips


some extra notes:
can we enforce a formatter for the whole package? it's hard to remember to not run `black .` and only format specific files

also the packaging and development setup is painful

.gitignore
Dreamsorcerer, line 1: https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/

aiocache/backends/valkey.py
```diff
@@ -34,18 +39,11 @@ class RedisBackend(BaseCache[str]):
...
         **kwargs,
     ):
         super().__init__(**kwargs)
 
-        # NOTE: decoding can't be controlled on API level after switching to
-        # redis, we need to disable decoding on global/connection level
-        # (decode_responses=False), because some of the values are saved as
-        # bytes directly, like pickle serialized values, which may raise an
-        # exception when decoded with 'utf-8'.
-        if client.connection_pool.connection_kwargs['decode_responses']:
```
Dreamsorcerer, line 47: Is there no way to get this info in glide? I'd assume the same issue would still be present.

aiocache/backends/valkey.py
```diff
@@ -126,16 +132,43 @@ async def _expire(self, key, ttl, _conn=None):
...
     async def _clear(self, namespace=None, _conn=None):
         if namespace:
-            keys = await self.client.keys("{}:*".format(namespace))
+            cursor, keys = await self.client.scan(b"0", "{}:*".format(namespace))
             if keys:
-                await self.client.delete(*keys)
+                return bool(await self.client.delete(keys))
         else:
-            await self.client.flushdb()
-        return True
```
Dreamsorcerer, line 138: This return has gone missing. if namespace and not keys would miss any return statements.

aiocache/backends/valkey.py
```diff
@@ -205,5 +240,4 @@ def parse_uri_path(cls, path):
         return options
 
     def __repr__(self):  # pragma: no cover
-        connection_kwargs = self.client.connection_pool.connection_kwargs
-        return "RedisCache ({}:{})".format(connection_kwargs['host'], connection_kwargs['port'])
```
Dreamsorcerer, line 209: None of this information available either in glide?

aiocache/lock.py
```diff
@@ -110,8 +116,15 @@ class OptimisticLock(Generic[CacheKeyType]):
     the one we retrieved when the lock started.
 
     Example usage::
+        from aiocache import ValkeyCache
+
+        from glide import GlideClient, GlideClientConfiguration, NodeAddress
+
+        addresses = [NodeAddress("localhost", 6379)]
+        conf = GlideClientConfiguration(addresses=addresses, database_id=0)
+        client = await GlideClient.create(conf)
```
Dreamsorcerer, line 125: This is going to be an issue for our decorators. Is it possible to create a client object without an async call (i.e. without creating the connections), and then starting that client using `async with`?

setup.cfg
```diff
@@ -7,6 +7,7 @@ max-line-length=100
 [tool:pytest]
 addopts = --cov=aiocache --cov=tests/ --cov-report term --strict-markers
 asyncio_mode = auto
+asyncio_default_fixture_loop_scope = function
```
Dreamsorcerer, line 10: Why did this need to change?

Dreamsorcerer:
Looking good. A bit of cleanup still to do and the tests need to pass, but looks like it's mostly good.

aiocache/backends/valkey.py
asafpamzn: In glide the data is not encoded, it is a raw/binary data. If you need to encode you need to call to the function encode.

for example
```
 value = get_random_string(10)

        res = await glide_client.set(
            key, value)
    
        assert await glide_client.get(key) == value.encode()
```

Does it answer your question?

Dreamsorcerer: OK, I thought it was a property of  the server connection, not of the library. So I guess this state is not possible then.

aiocache/backends/valkey.py
asafpamzn: 1. We can add it in the next version. 
2. What do we want to print? the primary or the replicas?

We can do an ugly patch:
```
client.config.addresses[0].host
client.config.addresses[0].port
```
If needed we can prioritize this support it is few minutes of work to add it.

Dreamsorcerer: I'm just thinking if someone has 2 clients to different servers for some reason, the repr should be able to distinguish them. I'll leave it to you to decide which host is exposed though.

aiocache/lock.py
Dreamsorcerer: Sure, we can probably do something like that for now.

I wouldn't describe what I'm asking for as lazy connect though, just moving the connection creation to `__aenter__()`. That way there is more control over when the connection is started.

Although, it probably makes sense to bring the client creation back into the Cache class. We moved the Redis client out of the Cache class, because it took so many different arguments, it was too awkward to maintain correctly. Here, we'll only ever need to pass a config object...

Dreamsorcerer: Yeah, let's just create the client inside the `GlideCache.__aenter__`. That's going to be a lot easier.

amirreza8002:
hi! I'm back

so currently i fixed set to return True or False
i replaced lua scripts with python code (also discovered that one of the lua scripts was probably broken)
fixed the CI image name
and some small cleanups

more importatnly tho, i looke into why tests are failing
what i discovered is that when we run the tests one by one, they work and pass
but when i run them in one place, they fail
it's probably a problem with how `pytest-asyncio` is configured
i'd have to look more into it, meanwhile i'd appreciate any input on the matter

on the discussion about context managers, i have something like this in mind


```py
async with ValkeyCache(config=my_config) as cache:
    await cache.set(...)
```

where `my_cache` is an instance of `GlideClientConfiguration` which user provides

another idea i have is to use a classmethod

```py
async with ValkeyCache.with_context_manager(config=congi):
    ...
```
this way we don't need to change the signature of `__init__`
but it's not a big deal anyway

but i do thing config should be provided by the user, so they can configure it as they please

Dreamsorcerer:
> this way we don't need to change the signature of `__init__`

There is no released version with the current signature. The first version is fine.

Dreamsorcerer:
> i'd have to look more into it, meanwhile i'd appreciate any input on the matter

Once the tests actually run in the CI, I might be able to help.

> p.s: i know the commits have gotten a bit messy, i'm keeping it like this so i can see diff, i'll clean it up before un-drafting

Don't bother, it makes reviewing more difficult. We'll squash commit the PR.

Dreamsorcerer:
Ah, that's awkward, there's also no pypy version for valkey-glide?

amirreza8002:
> Ah, that's awkward, there's also no pypy version for valkey-glide?

https://github.com/valkey-io/valkey-glide/blob/main/python/pyproject.toml#L24
this says that it does
but i don't know, haven't tested it

Dreamsorcerer:
```
ERROR    asyncio:base_events.py:1871 Task was destroyed but it is pending!
task: <Task cancelling name='Task-29' coro=<BaseClient._reader_loop() done, defined at /opt/hostedtoolcache/Python/3.13.2/x64/lib/python3.13/site-packages/glide/glide_client.py:509> wait_for=<Future cancelled>>
```

Is that a bug in glide? From what I see at a glance, the _reader_loop() task is never awaited on. It looks like it's been cancelled here, which appears to only happen when the client is garbage collected. Either we're not closing the client correctly, or glide is depending on `__del__()` for cleanup (it should cleanup in close() instead and await the cancelled task).

Dreamsorcerer:
> > Ah, that's awkward, there's also no pypy version for valkey-glide?
> 
> https://github.com/valkey-io/valkey-glide/blob/main/python/pyproject.toml#L24 this says that it does but i don't know, haven't tested it

Oh, well, pip can't find it to install on the pypy job for some reason.

Dreamsorcerer:
> ```
> ERROR    asyncio:base_events.py:1871 Task was destroyed but it is pending!
> task: <Task cancelling name='Task-29' coro=<BaseClient._reader_loop() done, defined at /opt/hostedtoolcache/Python/3.13.2/x64/lib/python3.13/site-packages/glide/glide_client.py:509> wait_for=<Future cancelled>>
> ```
> 
> Is that a bug in glide? From what I see at a glance, the _reader_loop() task is never awaited on. It looks like it's been cancelled here, which appears to only happen when the client is garbage collected. Either we're not closing the client correctly, or glide is depending on `__del__()` for cleanup (it should cleanup in close() instead and await the cancelled task).

Because we fail on warnings, that may be all the errors occurring in the tests. If fixed, that may resolve most of them.

Dreamsorcerer:
That's also the reason the tests work when run individually. For a single test run, the garbage collection won't happen until after the test run has completed, so the tests end up passing.

amirreza8002:
> > > Ah, that's awkward, there's also no pypy version for valkey-glide?
> > 
> > 
> > https://github.com/valkey-io/valkey-glide/blob/main/python/pyproject.toml#L24 this says that it does but i don't know, haven't tested it
> 
> Oh, well, pip can't find it to install on the pypy job for some reason.

i did try to install it locally
it did install, tho it's broken

amirreza8002:
> ```
> ERROR    asyncio:base_events.py:1871 Task was destroyed but it is pending!
> task: <Task cancelling name='Task-29' coro=<BaseClient._reader_loop() done, defined at /opt/hostedtoolcache/Python/3.13.2/x64/lib/python3.13/site-packages/glide/glide_client.py:509> wait_for=<Future cancelled>>
> ```
> 
> Is that a bug in glide? From what I see at a glance, the _reader_loop() task is never awaited on. It looks like it's been cancelled here, which appears to only happen when the client is garbage collected. Either we're not closing the client correctly, or glide is depending on `__del__()` for cleanup (it should cleanup in close() instead and await the cancelled task).

hmm, might be
i'll test it with different scopes, also i'll test it with anyio since they seem to handle tests a bit differently
if i find out it's breaking cause of our setup, i'll report back
otherwise i'll open an issue on glide's repo

amirreza8002:
so that fixed the problem...
imma ask you to forget what you see in that commit..

there are some ut problems, i'll  fix that soon and hopefully the work is finished!!

asafpamzn:
> the _reader_loop() task is never awaited on. It looks like it's been cancelled here, which appears to only happen when the client is garbage collected. Either we're not closing the client correctly, or glide is depending on `__del__()` for cleanup (it should cleanup in close() instead and await the cancelled task).

Might be,
@amirreza8002 , please let us know if you want me to help

Dreamsorcerer:
> > the _reader_loop() task is never awaited on. It looks like it's been cancelled here, which appears to only happen when the client is garbage collected. Either we're not closing the client correctly, or glide is depending on `__del__()` for cleanup (it should cleanup in close() instead and await the cancelled task).
> 
> Might be, @amirreza8002 , please let us know if you need my help

Looks like it just missed the close. The close() method should probably await the task though to ensure any exceptions etc. get raised correctly (like https://docs.aiohttp.org/en/stable/web_advanced.html#complex-applications).

amirreza8002:
hi @asafpamzn 
i've wondered about two things

one is the lack of a connection pool that we see in other clients
is this intentional? or something that's planned for future?

two os the lack of context manager support
also if it's intentional or something for future
thanks for allocating your time for this

amirreza8002:
hi again 👋🏼
so as far as i can tell this should do it

let me know what you think
if it's good, i should get to docs😮‍💨

aiocache/backends/valkey.py
```diff
@@ -194,10 +201,12 @@ class ValkeyCache(ValkeyBackend):
...
     def __init__(
         self,
-        client: GlideClient,
+        client: Optional[GlideClient] = None,
         serializer: Optional["BaseSerializer"] = None,
         namespace: str = "",
         key_builder: Callable[[str, str], str] = lambda k, ns: f"{ns}:{k}" if ns else k,
+        backend: type[GlideClient] = GlideClient,
+        config: GlideClientConfiguration = None,
         **kwargs: Any,
```
Dreamsorcerer, line 210: I find this a bit awkward and confusing. I think I'd rather make the config required and simplify it to:
```suggestion
        config: GlideClientConfiguration,
        serializer: Optional["BaseSerializer"] = None,
        *,
        namespace: str = "",
        key_builder: Callable[[str, str], str] = lambda k, ns: f"{ns}:{k}" if ns else k,
        **kwargs: Any,
```

asafpamzn:
> one is the lack of a connection pool that we see in other clients
> is this intentional? or something that's planned for future?

We will add it the future. But, connection pool is mainly used for sync API as each thread has to get a connection from the pool and return it once it got the response back from Valkey. For async mode it is less important as multiple tasks can share the same connection.
If it is blocking you we can add it. It will take sometime though.

> two os the lack of context manager support
> also if it's intentional or something for future
> thanks for allocating your time for this

I will work with the other maintainers to add it in the next release in end of May. 
I think that it should be straight forward and there should not be an objection. I will get back to you.

amirreza8002:
hi @asafpamzn , thanks for your response 🙌

about blockers, dreamsourcerer would have a better take

i did make a pr about something I'd seen here, but i wouldn't call that a blocker

we're currently discussing how to implement a context manager, i agree that the current implementation is ugly 
but i need some good sleep before working on it anymore 
in the mean time, if anyone has any ideas on it, I'd appreciate it

ikolomi:
@amirreza8002 @Dreamsorcerer 
Hi Guys,
I am also Glide's maintainer, nice to meet.

I will be looking into the issue with not tearing down the client object correctly, we might have a bug. Will update

Dreamsorcerer:
> about blockers, dreamsourcerer would have a better take

Main thing I'd like looked at ASAP is Pypy. If it's supposed to be supported, why can't the CI install it?

ikolomi:
@amirreza8002 @Dreamsorcerer 
I have looked into the issue with the clean up. Our code does require an explicit call to close() method as presented in out examples: [                await client.close()](https://github.com/valkey-io/valkey-glide/blob/main/examples/python/cluster_example.py#L120)


Having said that, we will look into an option to integrate the close() logic into the destructor. I am pretty sure we've already tried but will re-investigate

amirreza8002:
cleaned up some stuff 
and added more info to the repr message

btw did my commecnts on context manager come through?

.idea/.gitignore
Dreamsorcerer, line 1: I don't think you managed to setup that global gitignore correctly.

Dreamsorcerer:
> btw did my comments on context manager come through?

Which ones?

amirreza8002:
i left two comments on your review on context manager.
I'm not sure why they are not visible to you,
I'll repost them here

amirreza8002:
first comment:

so i did try that, there is one issue tho:
how to instantiate a client without a context manager?

since glide clients are made by awaiting the create() method, we can't do that in __init__
so it would become something like this:

```py
class ValkeyCache(...):
    def __init__(self, config, ...):
        self.client = await GlideClient.create(config=config)  # error
        
    async def __aenter__(self):
        return self
    
    async def __aexit__(self, *args, **kwargs):
        await self.client.close()
```

or is there something i'm missing?

amirreza8002:
i have two designs in mind

```py
class ValkeyCache(...):
    def __init__(self, config, ...):
        self.config = config
        
    async def __aenter__(self):
        await connect()
        return self
    
    async def __aexit__(self, *args, **kwargs):
        await self.client.close()
        
    async def connect(self):
        self.client = await GlideClient.create(config=self.config)
```

and document that the user should await on connect after instantiation

or:

```py
class ValkeyCache(...):
    def __init__(self, config, ...):
        self.config = config
        
    async def __aenter__(self):
        self.client = await GlideClient.create(config=self.config)
        return self
    
    async def __aexit__(self, *args, **kwargs):
        await self.client.close()

    @classmethod        
    async def connect(cls, config):
        client = await GlideClient.create(config=config)
        self = cls(config=config)
        self.client = client
        return self
```

admittedly both are somewhat ugly

Dreamsorcerer:
> I'm not sure why they are not visible to you,

Maybe you started a review and didn't submit the review?

> since glide clients are made by awaiting the create() method, we can't do that in **init**
> so it would become something like this:

Just create the client in `__aenter__()`. There's no need for it before that.

> how to instantiate a client without a context manager?

You mean a Cache? I don't think we need another way to do so. The user can just call `__aenter__()` if needed. Or if we really wanted to support a separate method, we could have start()/close(), but I'm not sure that's something we need yet.

Dreamsorcerer:
@asafpamzn @ikolomi Could one of you confirm the situation with Pypy support?

amirreza8002:
we discussed pypy here : https://github.com/valkey-io/valkey-glide/issues/3553
i think they are working on it


is the general feeling around this PR good?
should i start documenting the backend?

amirreza8002:
hi @Dreamsorcerer 

the glide team have informed me that pypy issue is fixed 
but the release will come in may

Dreamsorcerer:
Yeah, that's only a couple of weeks, so maybe we wait for that and then cleanup this PR.

amirreza8002:
hi again 🙌🏼
seems like they released the new version
i updated the dependencies, lets see if it works as expected

amirreza8002:
hi @ikolomi @asafpamzn 
can you see if the build error we're getting here is a dependency problem?
i don't see any dependencies mentioned in glide docs, if it's a dependency problem maybe we should add it to the docs

but even when i build it in my own system, in a pypy env, it doesn't seem to work, i'll open an issue for this one

asafpamzn:
> hi @ikolomi @asafpamzn can you see if the build error we're getting here is a dependency problem? i don't see any dependencies mentioned in glide docs, if it's a dependency problem maybe we should add it to the docs
> 
> but even when i build it in my own system, in a pypy env, it doesn't seem to work, i'll open an issue for this one
sure, we are on it

ikolomi:
> hi @ikolomi @asafpamzn can you see if the build error we're getting here is a dependency problem? i don't see any dependencies mentioned in glide docs, if it's a dependency problem maybe we should add it to the docs
> 
> but even when i build it in my own system, in a pypy env, it doesn't seem to work, i'll open an issue for this one

@amirreza8002 
It seems you dont have the protoc installed on the build system. Github runners dont include it i believe.
We install it explicitly in our build [workflow](https://github.com/valkey-io/valkey-glide/blob/77388529dd8662a6f689ed107fe564a5b5e94a38/.github/workflows/pypi-cd.yml#L165) 

Could you please verify

Dreamsorcerer:
> It seems you dont have the protoc installed on the build system.

So, there's no prebuilt wheels for pypy then? Would probably very useful to have those from a user perspective.

Looks like there's a libprotoc-dev package that can be apt installed, if that resolves the problem for now.

amirreza8002:
> We install it explicitly in our build workflow

ah I didn't see that part
tho i do think thos should be included in the documentation

thanks for checking this out 🙏

i reopened the issue i had on glide's repo, since it seems to persist
I'll test again once the issue is resolved

ikolomi:
@amirreza8002 @Dreamsorcerer 

We have just released 1.3.5 version which includes PyPy wheels. Previous release contained the sources which pip tried to build on you systems.

Appreciate if you will give it a try:
https://github.com/valkey-io/valkey-glide/issues/3553

requirements.txt
```diff
@@ -8,4 +8,4 @@ pytest==8.3.5
 pytest-asyncio==0.26.0
 pytest-cov==6.0.0
 pytest-mock==3.14.0
-redis==5.2.1
+valkey-glide==1.3.3
```
Dreamsorcerer, line 11:
```suggestion
valkey-glide==1.3.5
```

amirreza8002:
ah
i did not remember those :p

i'll hopefully get to it today or tomorrow

requirements.txt
amirreza8002: should we change it in setup.py as well?

amirreza8002:
so
mm
a question
how are we suppose to pass a cache instance to decorators? such as https://github.com/aio-libs/aiocache/blob/master/examples/multicached_decorator.py#L18

since we only can instantiate the cache client via async context managers, and i don't think that's an option here

in decorator tests such as https://github.com/aio-libs/aiocache/blob/master/tests/acceptance/test_decorators.py#L24-L25 we use pytest fixtures to create the object

but that's not an option here

i can't manually instantiate either since glide has to be instantiated by `await`ing

requirements.txt
Dreamsorcerer: If you like, but I assume any recent version should work as long as someone can install it.

Dreamsorcerer:
> i can't manually instantiate either since glide has to be instantiated by `await`ing

We already said that we'd instantiate valkey in the aenter method of the cache.

Dreamsorcerer:
> how are we suppose to pass a cache instance to decorators? such as https://github.com/aio-libs/aiocache/blob/master/examples/multicached_decorator.py#L18

That looks to me like it should work just fine, with an `async with` replacing the `cache.close()` inside the test function.

amirreza8002:
> > how are we suppose to pass a cache instance to decorators? such as https://github.com/aio-libs/aiocache/blob/master/examples/multicached_decorator.py#L18
> 
> That looks to me like it should work just fine, with an `async with` replacing the `cache.close()` inside the test function.

in the example the `cache` object is made by this https://github.com/aio-libs/aiocache/blob/master/examples/multicached_decorator.py#L15

but we can't do that here, since it has to be made via a async context manager

normally we do something like this:
```python
    async with ValkeyCache(config=config, namespace="main") as cache:
        await cache.delete("key")
        # or
        @multi_cached(cache, ...)
        async def inner_function(...):
            .
```
we could do something like 
```python
cache = await ValkeyCache(config=config, namespace="main")._connect()
```

but both scenarios would need to be inside an async function

with a decorator, we normally don't have an async function wrapping the decorated function

Dreamsorcerer:
> but we can't do that here, since it has to be made via an async context manager

No it doesn't...?

Dreamsorcerer:
Again, the only change I see needed is replacing `cache.close()` with `async with cache`.

amirreza8002:
i'm getting confused here
imma provide as much info as i can, hoefully it'll clear things up



> Again, the only change I see needed is replacing `cache.close()` with `async with cache`.

yes the context manager does close the connection when exiting
but it also [creates the connection](https://github.com/aio-libs/aiocache/pull/975/files#diff-303bf896e209e294a018c54407da672aaa2b25463c330aeb4e8cf047154798e0R40) when entering
and that is because glide has an [async only](https://github.com/aio-libs/aiocache/pull/975/files#diff-303bf896e209e294a018c54407da672aaa2b25463c330aeb4e8cf047154798e0R47) interface for connecting to the server

so even tho with redis you could do something like this
`RedisCache(...)` 
and get a working cache connected to a server

with valkey you can not, and 
`ValkeyCache(...)` is basically useless

and as far as i know, for decorators to work , they need to receive a cache object that is connected to a server

in order for us to have a working `ValkeyCache` object, we can either use it in a context manager
or await `._connect()`


so unless there's something i can't see, there is no way to create a working cache object and pass it to decorators out side of an async function

amirreza8002:
perhaps this commit clarifies more

Dreamsorcerer:
> and as far as i know, for decorators to work , they need to receive a cache object that is connected to a server

If that is true, then something is wrong with the design. I don't think you should have anything connected to a server when defining the functions.

Again, I'm expecting it to work with only making the changes to the test function:
```
+     async with cache:
....
-      await cache.close()
```

amirreza8002:
> Again, I'm expecting it to work with only making the changes to the test function


it doesn't seem to work, as my last commit might show


perhaps we can do

```python
async def _set(...):
    if not self.client:
        self.client = await self._connect()
```

tho it's not just _set, it should be on all cache methods 



let me know what you think 
hopefully I'll get to it when i wake up

Dreamsorcerer:
> it doesn't seem to work, as my last commit might show

Your last commit creates 2 caches? The first cache is never started..

examples/multicached_decorator.py
```diff
@@ -30,16 +27,16 @@ async def test_multi_cached():
     await multi_cached_ids(ids=("a", "c"))
     await multi_cached_keys(keys=("d",))
 
-    assert await cache.exists("a")
-    assert await cache.exists("b")
-    assert await cache.exists("c")
-    assert await cache.exists("d")
+    async with ValkeyCache(config=config, namespace="main") as cache:
```
Dreamsorcerer, line 30:
```suggestion
    async with cache:
```

Dreamsorcerer:
cache_decorator.py is already doing that correctly (not sure why multicached_decorator.py wasn't updated previously), just needs the config update.

amirreza8002:
the change to the makefile was done based on this doc
https://github.com/Arkq/flake8-requirements?tab=readme-ov-file#customization

to fix https://github.com/Arkq/flake8-requirements?tab=readme-ov-file#faq

tests/performance/test_footprint.py
```diff
@@ -1,77 +1,79 @@
...
-async def redis_client() -> AsyncIterator["redis.Redis[str]"]:
-    async with cast("redis.Redis[str]",
-                    redis.Redis(host="127.0.0.1", port=6379, max_connections=1)) as r:
-        yield r
+async def valkey_client() -> AsyncIterator["GlideClient"]:
+    addresses = [NodeAddress("localhost", 6379)]
+    conf = GlideClientConfiguration(addresses=addresses)
+    client = await GlideClient.create(conf)
+
+    yield client
```
Dreamsorcerer, line 16: I think this is breaking the tests, doesn't look like we're closing it?
```suggestion
    async with GlideClient.create(conf) as client:
        yield client
```

Dreamsorcerer:
Awesome, all tests are passing now. Just a few lines of lost coverage to resolve (I think most of the missed lines can probably be removed).

Dreamsorcerer:
Just realised that coverage uploads had been failing for the last year. I've cleaned up that up now, so diff actually makes sense.

There's still 3 lines which have lost coverage, and I'm unclear why:
https://app.codecov.io/gh/aio-libs/aiocache/pull/975/indirect-changes

Dreamsorcerer:
Ah, having removed that dodgy xfail line, it appears those tests are failing completely. That's probably the coverage loss.

aiocache/backends/valkey.py
```diff
@@ -167,21 +210,19 @@ class RedisCache(RedisBackend):
...
-    NAME = "redis"
+    NAME = "valkey"
 
     def __init__(
         self,
-        client: redis.Redis,
         serializer: Optional["BaseSerializer"] = None,
         namespace: str = "",
         key_builder: Callable[[str, str], str] = lambda k, ns: f"{ns}:{k}" if ns else k,
         **kwargs: Any,
```
Dreamsorcerer, line 223: Probably want to fix this. This is why the failing test wasn't caught by mypy. We could use a TypedDict or something here.

Dreamsorcerer: I'll do it in a separate PR.

tests/performance/server.py
```diff
@@ -57,9 +66,17 @@ async def handler_get(req: web.Request) -> web.Response:
     return web.Response(text=str(data))
 
 
+def cache_manager_ctx(backend: str) -> Callable[[web.Application], AsyncIterator[None]]:
+    async def ctx(app: web.Application) -> AsyncIterator[None]:
+        async with CacheManager(backend) as cm
```
Dreamsorcerer, line 71:
```suggestion
        async with CacheManager(backend) as cm:
```

Dreamsorcerer:
@asafpamzn @ikolomi It appears that Glide is segfaulting in the concurrency test (this is basically throwing 1500 requests at the server and checking that the majority of them successfully retrieve from the cache.

asafpamzn:
> @asafpamzn @ikolomi It appears that Glide is segfaulting in the concurrency test (this is basically throwing 1500 requests at the server and checking that the majority of them successfully retrieve from the cache): https://github.com/aio-libs/aiocache/actions/runs/15100087813/job/42439878618?pr=975#step:7:36

I see that it happens only on Python 3.12. We are on it

asafpamzn:
@Dreamsorcerer , can we execute the tests locally on my machine?
Should I create a  valkey-server? Is there any requirement for the machine type?

The tests basically creates 1500 processes of aiocache, each executes?

```
def run_server(backend: str) -> None:
    app = web.Application()
    app[cache_key] = CacheManager(backend)
    app.on_shutdown.append(app[cache_key].close)
    app.router.add_route("GET", "/", handler_get)
    web.run_app(app)
 ```

**Comment by Dreamsorcerer** [comment]
> I see that it happens only on Python 3.12. We are on it

No, it fails on all versions.

> The tests basically creates 1500 processes of aiocache, each executes?

Not quite. It creates a single server (web application) and then uses Apache Benchmark to send 1500 requests (I think 500 simultaneous) to that server.

I've not run them locally myself, but you should be able to run them. I'd assume a command like `pytest 'tests/performance/test_concurrency.py::test_concurrency_error_rates[valkey]'` would work (you'll need to have the ab command installed though).

**Comment by amirreza8002** [comment]
> pytest 'tests/performance/test_concurrency.py::test_concurrency_error_rates[valkey]'

I've not used it like this, might work
if not, `pytest tests/performance/test_concurrency.py::test_concurrency_error_rates -k valkey` will work

**Comment by asafpamzn** [comment]
I recreated locally on my mac. 
It fails when the concurrency is above 200. 
Debugging.
For some reason in my mac I don't see the stack trace of GLIDE

**Comment by Dreamsorcerer** [comment]
> For some reason in my mac I don't see the stack trace of GLIDE.

It segfaulted, so there is no stack trace. I assume you'd need to use a C-level debugger to get a better idea what is failing.

**Comment by amirreza8002** [comment]
i'm not getting any errors while running the test
it passes

maybe your problem in mac is related to this comment
```
    # On some platforms, it's required to enlarge number of "open file descriptors"
    #  with "ulimit -n number" before doing the benchmark.
```

i don't know why github is segfaulting

**Comment by amirreza8002** [comment]
sorry, misclick

**Comment by Dreamsorcerer** [comment]
> maybe your problem in mac is related to this comment

Can't be, the tests are working with the redis client (although the error rate increased in the last year, so now only 1-30% succeed, which is why someone added the xfail).

**Comment by amirreza8002** [comment]
> Can't be, the tests are working with the redis client (although the error rate increased in the last year, so now only 1-30% succeed, which is why someone added the xfail).

i meant the problem asafpamzan has with 200 concurencies

i really don't know much more
just that the test passes on my system

if more info helps:
i run valkey with `valkey --save ""`
my cpu is Intel(R) Xeon(R) W-2145 @ 4.50 GHz with 16 cores

if you need more, let me know

**Comment by asafpamzn** [comment]
Ack, thanks on it

**Comment by asafpamzn** [comment]
I did not manage to execute on my mac.
I am using ubuntu now and it works.
I tried with concurrency of 1000 and 60000 requests and it worked

```
 total_requests = 60000

    # On some platforms, it's required to enlarge number of "open file descriptors"

    #  with "ulimit -n number" before doing the benchmark.

    cmd = ("ab", "-n", str(total_requests), "-c", "1000", "http://127.0.0.1:8080/")

    result = subprocess.run(cmd, capture_output=True, check=True, encoding="utf-8")
```


This is the result

```
_________________________________ coverage: platform linux, python 3.12.3-final-0 _________________________________

Name                                    Stmts   Miss Branch BrPart  Cover
-------------------------------------------------------------------------
aiocache/__init__.py                       23      4      0      0    83%
aiocache/backends/__init__.py               0      0      0      0   100%
aiocache/backends/memcached.py             99     71     22      0    23%
aiocache/backends/memory.py                86     60     30      0    22%
aiocache/backends/valkey.py               145     82     48      3    34%
aiocache/base.py                          275     93     12      4    66%
aiocache/decorators.py                    137    111     38      0    15%
aiocache/exceptions.py                      2      2      0      0     0%
aiocache/lock.py                           52     33      4      0    34%
aiocache/plugins.py                        47     47     16      0     0%
aiocache/serializers/__init__.py           10      2      0      0    80%
aiocache/serializers/serializers.py        66     24      8      1    58%
tests/__init__.py                           0      0      0      0   100%
tests/acceptance/__init__.py                0      0      0      0   100%
tests/acceptance/conftest.py               24     24      0      0     0%
tests/acceptance/test_base.py             190    190      0      0     0%
tests/acceptance/test_decorators.py       141    141      4      0     0%
tests/acceptance/test_lock.py             209    209      2      0     0%
tests/acceptance/test_plugins.py           43     43      6      0     0%
tests/acceptance/test_serializers.py      112    112      0      0     0%
tests/conftest.py                           7      4      0      0    43%
tests/performance/__init__.py               0      0      0      0   100%
tests/performance/conftest.py              11      6      0      0    45%
tests/performance/server.py                70     14     10      3    74%
tests/performance/test_concurrency.py      27      0      0      0   100%
tests/performance/test_footprint.py       110    110     26      0     0%
tests/ut/__init__.py                        0      0      0      0   100%
tests/ut/backends/__init__.py               0      0      0      0   100%
tests/ut/backends/test_memcached.py       192    192      2      0     0%
tests/ut/backends/test_memory.py          158    158      0      0     0%
tests/ut/backends/test_valkey.py          155    155      2      0     0%
tests/ut/conftest.py                       36     36      2      0     0%
tests/ut/test_base.py                     414    414      0      0     0%
tests/ut/test_decorators.py               332    332      2      0     0%
tests/ut/test_exceptions.py                 3      3      0      0     0%
tests/ut/test_lock.py                      87     87      0      0     0%
tests/ut/test_plugins.py                   58     58      4      0     0%
tests/ut/test_serializers.py              118    118      0      0     0%
tests/utils.py                             47     47      2      0     0%
-------------------------------------------------------------------------
TOTAL                                    3486   2982    240     11    14%
======================================== 1 passed, 1 deselected in 21.46s ====
```

**Comment by BoazBD** [comment]
Hey folks,
I’m working on fixing that CI failure but can’t reproduce the NoScriptError locally with Python 3.11.11. Could you share the exact steps or commands I need to run to trigger it on my machine? Any pointers would be a huge help :)

**Comment by Dreamsorcerer** [comment]
All of those test runs have the same traceback too, so it's always the same place.

**Comment by BoazBD** [comment]
Hi,
I was able to reproduce the CI failure in my fork.
I investigated it a bit and discovered that starting Glide as a child process could cause a known asyncio [issue](https://github.com/python/cpython/issues/66197).
It isn’t a Glide bug, but how Python’s asyncio handles child processes.
To fix it, I ran the server using subprocess.Popen instead of running it as a child process, and it fixed the error in the CI.
Here’s the updated fixture:

```
# TODO: Fix and readd "memcached" (currently fails >98% of requests)
@pytest.fixture(params=("memory", "valkey"))
def server(request):
    cmd = [
        sys.executable, "-c",
        f"from tests.performance.server import run_server; run_server('{request.param}')"
    ]
    p = subprocess.Popen(cmd)
    time.sleep(2)
    yield
    p.terminate()
    try:
        p.wait(timeout=5)
    except subprocess.TimeoutExpired:
        p.kill()
        p.wait()
```
You can see the passing CI tests here: https://github.com/BoazBD/aiocache/actions/runs/15276932475/job/42965586715
Please let me know if this looks like a good solution, or if you have any other suggestions! :)

**Comment by Dreamsorcerer** [comment]
> I investigated it a bit and discovered that starting Glide as a child process could cause a known asyncio [issue](https://github.com/python/cpython/issues/66197).

Thanks for that. Weird that the other backends never had a problem with that.

That test is now passing on all Python versions (which also means that Glide has significantly improved the success rate over the redis client).

File: aiocache/backends/valkey.py [hunk 12]
```diff
@@ -58,59 +59,70 @@ async def _gets(self, key, encoding="utf-8", _conn=None):
...
+            kwargs["expiry"] = ExpirySet(ExpiryType.SEC, ttl)
         was_set = await self.client.set(key, value, **kwargs)
-        if not was_set:
+        if was_set != "OK":
             raise ValueError("Key {} already exists, use .set to update the value".format(key))
         return was_set
 
     async def _exists(self, key, _conn=None):
+        if isinstance(key, str):
+            key = [key]
```
Dreamsorcerer, line 125: Last check on the coverage.
These checks appear to always be True in the tests, should we expect non-str keys here? If so, we should probably have a test that covers it.

aiocache/backends/valkey.py
```diff
@@ -147,15 +188,17 @@ async def _raw(self, command, *args, encoding="utf-8", _conn=None, **kwargs):
         return value
```
Dreamsorcerer, line 188: The line above (can't comment on it because it's not in the diff) is no longer covered (command is never keys or mget).

aiocache/backends/valkey.py
amirreza8002: yup
iirc glide only accepted lists there, so i added that to support str
i'll add a test

amirreza8002:
@Dreamsorcerer 
the `exists` method of glide is a bit different than the one one we have
https://github.com/valkey-io/valkey-glide/blob/main/python/python/glide/async_commands/core.py#L2604-L2629
glide's exists takes a list of keys and returns how many of those keys are avilable

but the way we handle it is by `bool()`https://github.com/aio-libs/aiocache/pull/975/files/4ca6c92f835ca9a264f3846fa6dd20761e7fadc4#diff-303bf896e209e294a018c54407da672aaa2b25463c330aeb4e8cf047154798e0R127
this works if we only check one key, and only get 0 or 1
but when we pass in multiple keys, true or false is not a good response

should i remove the `bool()`?
if so, the API will be different rom other backends that return a bool

aiocache/backends/valkey.py
Dreamsorcerer: Doesn't the current BaseCache API always expect a str? Maybe we should just keep it with a single str and just pass `[key]` unconditionally?

aiocache/backends/valkey.py
```diff
@@ -0,0 +1,251 @@
...
+            raise TypeError("Value is not an integer") from None
+
+    async def _expire(self, key, ttl, _conn=None):
+        if ttl == 0:
+            return await self.client.persist(key)
+        return await self.client.expire(key, ttl)
+
+    async def _delete(self, key, _conn=None):
+        if isinstance(key, str):
+            key = [key]
```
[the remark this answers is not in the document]
amirreza8002: is this one necessary?
we're not changing the result to bool here
so might as well allow a list of keys

amirreza8002:
finally 🎆
i'll work on it's documentation, tho as a non-english speaker i'd appreciate help on that 
enjoyed this work ❤️

asafpamzn:
@Dreamsorcerer , we will be happy to see it merged in. :) 
I guess that you are busy, please let me know if we can help

amirreza8002:
sorry I didn't get the chance to document this
I'll get to it soon

.gitignore
```diff
@@ -80,6 +80,7 @@ celerybeat-schedule
 .env
 
 # virtualenv
+.venv/
```
Dreamsorcerer, line 83:
```suggestion
```

Makefile
```diff
@@ -2,10 +2,9 @@ cov-report = true
 
 
 lint:
-	flake8 tests/ aiocache/
-
+	flake8 tests/ aiocache/ '--known-modules=valkey-glide:[glide]'
```
Dreamsorcerer, line 5:
```suggestion
	flake8 tests/ aiocache/
```

aiocache/backends/valkey.py
```diff
@@ -0,0 +1,250 @@
...
+    from typing import Self
+else:
+    from typing import Any as Self
+
+
+logger = logging.getLogger(__name__)
+
+
+class ValkeyBackend(BaseCache[str]):
+    def __init__(self, config: GlideClientConfiguration = None, **kwargs):
```
Dreamsorcerer, line 33: I still can't really imagine any reason for passing None here. I'm inclined to just make this required and simplify the rest.

aiocache/backends/valkey.py
```diff
@@ -0,0 +1,248 @@
...
+from glide import (
+    ConditionalChange,
+    ExpirySet,
+    ExpiryType,
+    GlideClient,
+    GlideClientConfiguration,
+    Script,
+    Transaction,
+)
+from glide.exceptions import RequestError as IncrbyException
```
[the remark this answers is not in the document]
amirreza8002: https://github.com/aio-libs/aiocache/blob/master/aiocache/backends/redis.py#L5
it was like this in redis backend, i kept the name

aiocache/backends/valkey.py
```diff
@@ -0,0 +1,248 @@
...
+        return self
+
+    async def __aexit__(self, *args, **kwargs) -> None:
+        await self._disconnect()
+
+    async def _connect(self, config: GlideClientConfiguration) -> GlideClient:
+        return await GlideClient.create(config=config)
+
+    async def _disconnect(self) -> None:
+        await self.client.close()
```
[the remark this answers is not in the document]
amirreza8002: i was thinking having a helper method would be better in case you decide to alter the API in future
i'll remove it if you want

aiocache/backends/valkey.py
```diff
@@ -0,0 +1,248 @@
...
+
+    async def _get(self, key, encoding="utf-8", _conn=None):
+        value = await self.client.get(key)
+        if encoding is None or value is None:
+            return value
+        return value.decode(encoding)
+
+    async def _gets(self, key, encoding="utf-8", _conn=None):
+        return await self._get(key, encoding=encoding, _conn=_conn)
+
```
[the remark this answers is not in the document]
amirreza8002: @Dreamsorcerer 
https://github.com/aio-libs/aiocache/blob/master/aiocache/lock.py#L146
locks seem to expect this method to exists

perhaps we do the following:
```python
_gets = get
```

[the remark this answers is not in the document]
Dreamsorcerer: Yes, sorry, it's an abstractmethod actually, so will need to keep this. The assignment might work though.

aiocache/backends/valkey.py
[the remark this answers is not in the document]
Dreamsorcerer: I don't see how it makes any difference. We can make this change later if we need to.

aiocache/lock.py
```diff
@@ -43,10 +43,15 @@ class RedLock(Generic[CacheKeyType]):
 
     Example usage::
 
-        from aiocache import Cache
+        from aiocache import ValkeyCache
         from aiocache.lock import RedLock
```
Dreamsorcerer, line 48:
```suggestion
```

tests/ut/backends/test_valkey.py
```diff
@@ -0,0 +1,256 @@
+import re
+from unittest.mock import ANY, AsyncMock, patch
```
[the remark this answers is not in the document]
amirreza8002: solved

tests/ut/backends/test_valkey.py
```diff
@@ -0,0 +1,256 @@
...
+    await valkey.__aexit__()
+
+
+class TestValkeyBackend:
+    async def test_backend_raise_if_no_backend(self):
+        msg = re.escape(
+            "ValkeyBackend.__init__() missing 1 required positional argument: 'config'"
+        )
+        with pytest.raises(TypeError, match=msg):
+            async with ValkeyBackend() as _:
```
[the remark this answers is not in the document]
Dreamsorcerer: https://github.com/aio-libs/aiocache/actions/runs/15542095789/job/43755166776

tests/performance/test_footprint.py
```diff
@@ -1,77 +1,81 @@
...
-import redis.asyncio as redis
+from glide import GlideClient, GlideClientConfiguration, NodeAddress
 
 
 @pytest.fixture
-async def redis_client() -> AsyncIterator["redis.Redis[str]"]:
-    async with cast("redis.Redis[str]",
-                    redis.Redis(host="127.0.0.1", port=6379, max_connections=1)) as r:
-        yield r
+async def valkey_client() -> AsyncIterator["GlideClient"]:
```
Dreamsorcerer, line 11:
```suggestion
async def valkey_client() -> AsyncIterator[GlideClient]:
```

tests/performance/test_footprint.py
Dreamsorcerer, line 1: This one's a bit odd. Why are we testing the redis/glide client...?

tests/ut/backends/test_valkey.py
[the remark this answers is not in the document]
Dreamsorcerer: Doesn't seem useful to me. The isn't testing our actual code, and is easily covered by type checkers.

tests/ut/backends/test_valkey.py
```diff
@@ -0,0 +1,256 @@
...
+
+    def test_inheritance(self, valkey_config):
+        assert isinstance(ValkeyCache(config=valkey_config), BaseCache)
+
+    def test_default_serializer(self, valkey_config):
+        assert isinstance(ValkeyCache(config=valkey_config).serializer, JsonSerializer)
+
+    @pytest.mark.parametrize(
+        "path,expected",
+        [("", {}), ("/", {}), ("/1", {"db": "1"}), ("/1/2/3", {"db": "1"})],
```
Dreamsorcerer, line 237:
```suggestion
        (("", {}), ("/", {}), ("/1", {"db": "1"}), ("/1/2/3", {"db": "1"})),
```

tests/ut/backends/test_valkey.py
```diff
@@ -0,0 +1,256 @@
...
+    def test_parse_uri_path(self, path, expected, valkey_config):
+        assert ValkeyCache(config=valkey_config).parse_uri_path(path) == expected
+
+    @pytest.mark.parametrize(
+        "namespace, expected",
+        (
+            [None, "test:" + ensure_key(Keys.KEY)],
+            ["", ensure_key(Keys.KEY)],
+            ["my_ns", "my_ns:" + ensure_key(Keys.KEY)],
+        ),  # noqa: B950
```
Dreamsorcerer, line 248:
```suggestion
            (None, "test:" + ensure_key(Keys.KEY)),
            ("", ensure_key(Keys.KEY)),
            ("my_ns", "my_ns:" + ensure_key(Keys.KEY)),
        ),
```

tests/ut/backends/test_valkey.py
```diff
@@ -0,0 +1,245 @@
+import re
```
Dreamsorcerer, line 1:
```suggestion
```

aiocache/backends/valkey.py
```diff
@@ -0,0 +1,202 @@
...
+        _, keys = await self.client.scan(b"0", "{}:*".format(namespace))
+        if keys:
+            return bool(await self.client.delete(keys))
+
+        return True
+
+    async def _raw(self, command, *args, encoding="utf-8", _conn=None, **kwargs):
+        value = await getattr(self.client, command)(*args, **kwargs)
+        if encoding is not None:
```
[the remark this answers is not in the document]
amirreza8002: sure i'll add one

tests/acceptance/test_base.py
```diff
@@ -213,40 +213,47 @@ async def test_close(self, memcached_cache):
...
 
-    async def test_clear_with_namespace_redis(self, redis_cache):
-        await redis_cache.set(Keys.KEY, "value", namespace="test")
-        await redis_cache.clear(namespace="test")
+    async def test_raw_no_encoding(self, valkey_cache):
+        await valkey_cache.set(Keys.KEY, "value")
+        # you can't pass `encoding` to raw()
+        assert (
+            await valkey_cache._raw("get", Keys.KEY, encoding=None) == "value".encode()
+        )
```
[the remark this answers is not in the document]
amirreza8002: hi
I'm not sure if the serializer can be None or not
I'll check if there are is anything restraining it

amirreza8002:
👀
I'm not sure what you mean by "this what happens..."
we're safe and sound 
just distracted by the news of our responses

sakosha:
Hey @amirreza8002! Hope you are safe!

Are you okay, if I finish this PR?

Dreamsorcerer:
Thanks for sticking with this PR, it turned into a much larger PR than I originally expected. That's been really helpful.
If anyone can create a followup PR for updating the docs, that would be much appreciated. No rush though.

Issue #684, opened 2023-03-04
Merge cache implementation classes for each backend


pshafer-als:
## Background

Each backend cache implementation is currently split into two classes, with the "interface" class derived from the "backend" class.

* `SimpleMemoryCache` subclasses `SimpleMemoryBackend`, which subclasses `BaseCache[str]`.
* `MemcachedCache` subclasses `MemcachedBackend`, which subclasses `BaseCache[bytes]`.
* `RedisCache` subclasses `RedisBackend`, which subclasses `BaseCache[str]`.

## Suggestion

**Merge each pair into a single subclass of `BaseClass`.** In other words, the `...Backend` classes do not appear to provide any benefit or value by being distinct classes.

* `SimpleMemoryCache` is highly bound to `SimpleMemoryBackend`, even in its name. It seems unlikely that anyone would use `SimpleMemoryBackend` directly, rather than using `SimpleMemoryCache`. 
* Also `SimpleMemoryBackend` does not appear to provide value as a base class for user customization. If a specialized memory cache is needed, one would probably derive from `SimpleMemoryCache` and override only what is needed, or derive from `BaseCache` if more extensive changes are required.

## Question

Are any `aiocache` users relying on `SimpleMemoryBackend`, `MemcachedBackend`, `RedisBackend` classes, or could their functionality be merged into the corresponding `...Cache` implementations?

sakosha:
Hello! Can I take this one?

Issue #609, opened 2022-12-31
Lifecycle management of decorators


Dreamsorcerer:
Cache objects should be closed at the end of an application lifecycle, with `await Cache.close()`, or using `async with`.

The current design of decorators creates a new cache instance with each decorator and no attempt to close it is made, thus failing to manage this lifecycle management properly.

e.g.
```
@cached(...)  # New cache is created here, but never closed.
def foo(): ...
```

We also want to consider using aiojobs for managing some background tasks, but this additionally requires being created within a running loop, something which is unlikely when a decorator is called.

----

One solution I can think of, is to explicitly manage the caches, and pass them to the decorators. This may also need a `.start()` method to initiate the cache later. e.g.

```
cache = Cache(...)

@cached(cache)
def foo(): ...

async def main():
    # Initialise application
    cache.start()
    # Run application
    ...
    # Cleanup application
    await cache.close()
```

Or more succinctly:
```
async def main():
    async with cache:
        # Run application
```

Dreamsorcerer:
These are causing warnings in our examples as well: #610.

cancan101:
>The current design of decorators creates a new cache instance with each decorator 

Is that true if an alias is used? ie if I use the same alias for multiple decorators will the underlying `Cache` instance be shared?

Dreamsorcerer:
> > The current design of decorators creates a new cache instance with each decorator
> 
> Is that true if an alias is used? ie if I use the same alias for multiple decorators will the underlying `Cache` instance be shared?

Yes, that should be shared, but the issue is really around the lifecycle of the cache, not the sharing. With an alias, the cache is still created on-demand and users are unlikely to think about closing the cache at any point.

I also don't see any real advantage to an alias like that, compared with just passing an explicit reference to the cache object directly.

mavwolverine:
Won't closing cache on application exit be a problem in an autoscaled environment?

elecnix:
Being able to close all caches would be nice.

With FastAPI:
    
    @app.on_event("shutdown")
    async def close_all_caches():
        for cache in aiocache.caches:
            try:
                await cache.close()
            except Exception:
                pass

Dreamsorcerer:
I think that's trivial for the user to track the caches and do that themselves anyway. Doing it at the library level would likely cause issues if other libraries used aiocache and didn't expect to have their cache closed by another library or application.

sakosha:
Hello @Dreamsorcerer!

[...]

Issue #608, opened 2022-12-31
Make caches Generic


Dreamsorcerer:
We should consider making `BaseCache` Generic, so we can provide better type safety when relevant.

Code could then look something like this, with mypy checking:
```
cache: Cache[str] = Cache(...)
await cache.get("foo")  # -> str
await cache.set("foo", "bar")  # OK
await cache.set("foo", 45)  # Error: Expected str
```

Existing typing behaviour can be reproduced by annotating it with `Cache[Any]`.

PR #984, opened 2025-04-26
Preserve type hints in decorator (#512)

### LINKED ISSUE #512: Typing

elecnix:
The @cached decorator did not preserve type hints, breaking static analysis.

## What do these changes do?

This PR fixes type hint preservation in the `@cached` decorator by using `ParamSpec` and `TypeVar` in its implementation.

It also adds a test to verify that type hints are correctly preserved on decorated functions.

No user-facing behavior changes, but static analysis and type checking for decorated functions now work as expected in pyright.

On the other hand, mypy does not correctly infer the return type of the decorated function. This appears to be a limitation with mypy.

## Related issue number

Fixes #512

## Checklist

### DIFF
--- a/aiocache/decorators.py
+++ b/aiocache/decorators.py
@@ -2,6 +2,10 @@ import asyncio
 import functools
 import inspect
 import logging
+from typing import Callable, TypeVar, ParamSpec, Coroutine, Any
+
+P = ParamSpec("P")
+R = TypeVar("R")
 
 from aiocache.base import SENTINEL
 from aiocache.lock import RedLock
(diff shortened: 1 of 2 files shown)


Dreamsorcerer:
The tasks under the milestone need to be implemented in roughly the order listed:
https://github.com/aio-libs/aiocache/milestone/8

Type annotations will not work correctly before then (we already enabled them in 0.12 and had to disable them again).

Dreamsorcerer:
Wait, Github has removed the ordering of milestone issues... that's rather annoying...

Dreamsorcerer:
Order may have been:

#684
#625
#671
#609
#573
#608
#610
#512

PR #985, opened 2025-04-26
Wrap decorators and add cache control API (#625)

### LINKED ISSUE #625: Wrap functions in class with decorator

elecnix:
## What do these changes do?

Refactored `@cached` and `@multi_cached` to return a wrapper object providing cache control:
- **refresh**: forces a cache update for the given key.
- **invalidate**: clears the cache for the given key, or clears all if no key is given.

Users can now call `refresh()` and `invalidate()` on decorated functions or methods to control the cache directly.

No breaking changes to existing cache usage patterns.

## Related issue number

Fixes #625
Replaces #927

## Checklist

### DIFF
--- a/aiocache/decorators.py
+++ b/aiocache/decorators.py
@@ -44,12 +44,38 @@ class cached:
         self.cache = cache
 
     def __call__(self, f):
-        @functools.wraps(f)
-        async def wrapper(*args, **kwargs):
-            return await self.decorator(f, *args, **kwargs)
+        class CachedFunctionWrapper:
+            def __init__(self, decorator, func):
+                functools.update_wrapper(self, func)
+                self._decorator = decorator
+                self._func = func
+                self.cache = decorator.cache
 
-        wrapper.cache = self.cache
-        return wrapper
+            async def __call__(self, *args, **kwargs):
+                return await self._decorator.decorator(self._func, *args, **kwargs)
+
+            async def refresh(self, *args, **kwargs):
+                """
+                Force a refresh of the cache.
+
+                This method recomputes the result by calling the original function,
+                then updates the cache with the new value for the given arguments.
+                """
+                return await self._decorator.decorator(
+                    self._func, *args, cache_read=False, cache_write=True, **kwargs
+                )
+
+            async def invalidate(self, *args, **kwargs):
+                """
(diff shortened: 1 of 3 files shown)


Dreamsorcerer:
Maybe this one is fine to go in now. I'll review later.

aiocache/decorators.py
```diff
@@ -44,12 +44,38 @@ def __init__(
         self.cache = cache
 
     def __call__(self, f):
-        @functools.wraps(f)
-        async def wrapper(*args, **kwargs):
-            return await self.decorator(f, *args, **kwargs)
+        class CachedFunctionWrapper:
```
Dreamsorcerer, line 47: This should be defined at the top level. We wouldn't be able to reference the type otherwise when we add type annotations.

aiocache/decorators.py
Dreamsorcerer, line 1: If you refer back to the example in the issue I posted, my intention was to replace the cached class completly. This currently feels just as awkward as before with the wrapper class referencing the decorator class.

My expectation is that the cached decorator will literally just be defined as:
```
def cached(func):
    return Wrapper(func)
```

All other logic can exist in the Wrapper class.

aiocache/decorators.py
```diff
@@ -44,12 +44,38 @@ def __init__(
...
+
+            async def invalidate(self, *args, **kwargs):
+                """
+                Invalidate the cache for the given key, or clear all cache if no key is provided.
+                """
+                if not args and not kwargs:
+                    await self.cache.clear()
+                else:
+                    key = self._decorator.get_cache_key(self._func, args, kwargs)
+                    await self.cache.delete(key)
```
Dreamsorcerer, line 76: Can we skip this in the initial PR and add it in a followup PR?

PR #986, opened 2025-04-26
feat(typing)!: make cache generic over value (#608)

elecnix:
## What do these changes do?

- Make BaseCache generic over both key and value types for improved type safety.
- Update all cache backends and usages to use BaseCache[KeyType, ValueType].
- Improve type annotations for cache methods.

BREAKING CHANGE: BaseCache and all backends now require two type parameters (key, value). Existing code using a single type parameter must be updated.

Before, users could write:

    cache: BaseCache[str] = ...

Users must now write:

    cache: BaseCache[str, Any] = ...
    # or for stricter typing:
    cache: BaseCache[str, int] = ...

## Related issue number

Fixes: #608

## Checklist

### DIFF
--- a/aiocache/backends/memcached.py
+++ b/aiocache/backends/memcached.py
@@ -1,5 +1,5 @@
 import asyncio
-from typing import Optional
+from typing import Any, Optional
 
 import aiomcache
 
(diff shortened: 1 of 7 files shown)


aiocache/backends/memcached.py
```diff
@@ -1,13 +1,13 @@
...
+from typing import Any, Optional
 
 import aiomcache
 
 from aiocache.base import BaseCache
 from aiocache.serializers import JsonSerializer
 
 
-class MemcachedBackend(BaseCache[bytes]):
+class MemcachedBackend(BaseCache[bytes, Any]):
```
Dreamsorcerer, line 10: I think all these need to be defined as a TypeVar. Then the user would need to annotate their code as: `MemcachedCache[Any]` and similar.

Though, probably better to get #684 done before this one.

PR #1011, opened 2025-06-17
Merge Caches with Backends

### LINKED ISSUE #684: Merge cache implementation classes for each backend

sakosha:
## What do these changes do?

* Merge `MemoryBackend` w/ `MemoryCache`.
* Merge `MemcachedBackend` w/ `MemcachedCache`.
* Merge `ValkeyBackend` w/ `ValkeyCache`.
* Update tests to use `...Cache`.
* Add type annotations for these classes.
* Add note to migrating docs.

* `Backend` classes removed.
* `typing-extensions` now a required dependency before 3.11

## Related issue number

Fixes #684

## Checklist

### DIFF
--- a/aiocache/backends/valkey.py
+++ b/aiocache/backends/valkey.py
@@ -3,12 +3,12 @@ import sys
 from typing import Any, Iterable, Literal, Mapping, Optional, TypedDict, Union, overload
 
 from glide import (
+    Batch,
     ConditionalChange,
     ExpirySet,
     ExpiryType,
     GlideClient,
     GlideClientConfiguration,
-    Transaction,
 )
 from glide.exceptions import RequestError as IncrbyException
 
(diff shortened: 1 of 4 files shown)


aiocache/backends/valkey.py
Dreamsorcerer, line 1: It'd be good to have the relevant changes separately from the typing changes...

sakosha: Should we also omit the **kwargs TODO?

Dreamsorcerer: Just leave it as is?

sakosha: Done!
see: #1024

PR #1019, opened 2025-06-21
Try memcached concurrency

### LINKED ISSUE #585: Concurrency test failures
The memcached test in tests/performance/test_concurrency.py is currently failing with >98% requests failing.

It would be great if anyone has some time to look into debugging this (just need to readd "memcached" to the fixture at the top of the file to start testing it again).

Another problem is when running these tests on Pypy, where only 1 request succeeds on the redis test for some reason (which probably suggests the initial set() call works, but none of the get() calls).

Dreamsorcerer:
Fixes #585.

### DIFF
--- a/tests/performance/test_concurrency.py
+++ b/tests/performance/test_concurrency.py
@@ -12,8 +12,7 @@ from .server import run_server
 set_start_method("spawn")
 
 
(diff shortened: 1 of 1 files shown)


Dreamsorcerer:
Must have been resolved by the spawn fix.

PR #1025, opened 2025-07-04
Tests to increase backend coverage

sakosha:
## What do these changes do?

Add tests for user-provided `serializer` arguments for backends.
Add test for user-provided `key_builder` argument for `ValkeyCache`

## Related issue number

Implements suggested comment: https://github.com/aio-libs/aiocache/pull/1024#issuecomment-3033201058

## Checklist

### DIFF
--- a/tests/ut/backends/test_memcached.py
+++ b/tests/ut/backends/test_memcached.py
@@ -258,7 +258,7 @@ class TestMemcachedCache:
     def test_build_key_no_spaces(self, memcached_cache):
         assert memcached_cache.build_key("hello world") == b"hello_world"
 
-    def test_custom_serializer(self, valkey_config):
+    def test_custom_serializer(self):
         assert isinstance(
             MemcachedCache(serializer=PickleSerializer()).serializer,
             PickleSerializer,
--- a/tests/ut/backends/test_memcached.py
+++ b/tests/ut/backends/test_memcached.py
@@ -3,7 +3,7 @@ from unittest.mock import AsyncMock, patch
 import aiomcache
 import pytest
 
-from aiocache.backends.memcached import MemcachedBackend, MemcachedCache
+from aiocache.backends.memcached import MemcachedCache
 from aiocache.base import BaseCache
 from aiocache.serializers import JsonSerializer
 from ...utils import Keys, ensure_key
@@ -11,7 +11,7 @@ from ...utils import Keys, ensure_key
 
 @pytest.fixture
 def memcached():
-    memcached = MemcachedBackend()
+    memcached = MemcachedCache()
     with patch.object(memcached, "client", autospec=True) as m:
         # Autospec messes up the signature on the decorated methods.
         for method in (
@@ -25,10 +25,16 @@ def memcached():
         yield memcached
 
 
-class TestMemcachedBackend:
+class TestMemcachedCache:
+    @pytest.fixture
+    def set_test_namespace(self, memcached_cache):
+        memcached_cache.namespace = "test"
+        yield
+        memcached_cache.namespace = None
+
     def test_setup(self):
         with patch.object(aiomcache, "Client", autospec=True) as aiomcache_client:
-            memcached = MemcachedBackend()
+            memcached = MemcachedCache()
 
             aiomcache_client.assert_called_with("127.0.0.1", 11211, pool_size=2)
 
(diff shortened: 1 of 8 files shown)


tests/ut/backends/test_memory.py
```diff
@@ -198,3 +198,9 @@ def test_default_serializer(self):
 
     def test_parse_uri_path(self):
         assert SimpleMemoryCache().parse_uri_path("/1/2/3") == {}
+
+    def test_custom_serializer(self, valkey_config):
```
Dreamsorcerer, line 202: This is test_memory.py. Probably the cause of the import failure.

```suggestion
    def test_custom_serializer(self):
```

sakosha: Good catch!

tests/ut/backends/test_valkey.py
```diff
@@ -240,3 +240,41 @@ def test_build_key_double_dot(
...
+        )
+
+    def test_default_key_builder(self, valkey_config):
+        real_code = ValkeyCache(config=valkey_config)._build_key.__code__
+        expected_code = (lambda k, ns: f"{ns}:{k}" if ns else k).__code__
+        assert (
+            real_code.co_code == expected_code.co_code
+            and real_code.co_consts == expected_code.co_consts
+            and real_code.co_names == expected_code.co_names
+        )
```
Dreamsorcerer, line 257: I feel like a cleaner way to test this would be to provide a different key builder and test a call produces the expected result.

e.g. `lambda k, ns: f"{ns}__k" if ns else k`
Then verify that using the cache with a namespace saves with the `{ns}__k` key.

tests/ut/backends/test_valkey.py
```diff
@@ -240,3 +240,41 @@ def test_build_key_double_dot(
 
     def test_build_key_no_namespace(self, valkey_cache):
         assert valkey_cache.build_key(Keys.KEY, namespace=None) == Keys.KEY
+
+    def test_custom_serializer(self, valkey_config):
+        assert isinstance(
+            ValkeyCache(config=valkey_config, serializer=PickleSerializer()).serializer,
+            PickleSerializer,
+        )
```
Dreamsorcerer, line 248: Ideally, it'd also be nice to actually exercise the serializer too, so we see it works all the way through (i.e. verify we can save and retrieve something).

tests/ut/backends/test_valkey.py
```diff
@@ -240,3 +240,37 @@ def test_build_key_double_dot(
...
+        async with ValkeyCache(
+            config=valkey_config,
+        ) as default:
+            await default.set(Keys.KEY.value, "value", namespace=namespace)
+
+        async with ValkeyCache(
+            config=valkey_config, key_builder=lambda k, ns: f"{ns}:{k}" if ns else k
+        ) as explicit:
+            assert await explicit.exists(Keys.KEY.value, namespace=namespace)
+            assert await explicit._exists(built_key)
```
Dreamsorcerer, line 262: I think I'd personally bypass any of the abstraction for this assertion:
```suggestion
            await default.set(Keys.KEY.value, "value", namespace=namespace)
            assert await default.client.exists([built_key])
```

tests/ut/backends/test_valkey.py
```diff
@@ -240,3 +240,37 @@ def test_build_key_double_dot(
...
+
+    async def test_custom_key_builder(self, valkey_config):
+        async with ValkeyCache(
+            config=valkey_config, key_builder=lambda k, ns: f"{ns}__{k}" if ns else k
+        ) as vc:
+            await vc.set(Keys.KEY, "value", namespace="namespace")
+            assert await vc.get(Keys.KEY, namespace="namespace") == "value"
```
Dreamsorcerer, line 281: And do much the same for the custom one:
```suggestion
            assert await vc.get(Keys.KEY, namespace="namespace") == "value"
            assert await vc.client.exists(["namespace__key"])
```

tests/ut/backends/test_valkey.py
sakosha: I'm okay with that, but could you please elaborate why are do you want to tight the test to implementation?

tests/ut/backends/test_memcached.py
```diff
@@ -257,3 +258,17 @@ def test_build_key_no_namespace(self, memcached_cache):
...
+        value = {"one": {"nested": "1"}, "two": 2}
+        serialized = pickle.dumps(
+            value,
+        )
+
+        async with MemcachedCache(serializer=PickleSerializer()) as mc:
+            assert isinstance(
+                mc.serializer,
+                PickleSerializer,
+            )
```
Dreamsorcerer, line 272:
```suggestion
        serialized = pickle.dumps(value)

        async with MemcachedCache(serializer=PickleSerializer()) as mc:
            assert isinstance(mc.serializer, PickleSerializer)
```

Dreamsorcerer:
Thanks for that.

PR #1031, opened 2025-07-16
add msgspec

Vizonex:
## What do these changes do?

I saw that this library was using msgspack so I felt like adding something of my own as well. Msgspec is one of the fastest python dataclass libraries out there because it's written in C and with it having it's own msgspack protocol I decided to add in an encoder and decoder.

I attempeted to make sure that my end was properly typehinted. 
I'm going to convert this into a draft for now but hopefully maybe I can turn it into a real pull request.

## Checklist

### DIFF
--- a/aiocache/__init__.py
+++ b/aiocache/__init__.py
@@ -30,7 +30,11 @@ else:
     _AIOCACHE_CACHES.append(MemcachedCache)
     del aiomcache
 
-from .decorators import cached, cached_stampede, multi_cached  # noqa: E402,I202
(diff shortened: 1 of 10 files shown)


Dreamsorcerer:
I'm not sure we want to be adding more dependencies here. As per other requests for new backends, we should probably put any other serializer in a new library (or a component in the parent library, if they're interested) and have it maintained there.

At the same time, a user wanting this could just implement the custom serializer in a couple of lines of code..

PR #1080, opened 2026-07-16
Guard against re-entrant ValkeyCache context use

### LINKED ISSUE #1023: Potential memory leak/bug in the valkey client
I think there might be an issue with the `__aenter__` method on the ValkeyBackend. 
https://github.com/aio-libs/aiocache/blob/d685fff3303138eb466a69bb5dbb4d6a345ef1ae/aiocache/backends/valkey.py#L32-L37

This code snippet explains my thinking. 
```
async def cache_operations():
    client_1 = None
    async with ValkeyCache(
        config=GlideClientConfiguration(addresses=[NodeAddress("localhost", 6379)])
    ) as my_cache_1: # <<< new client created
        client_1 = my_cache_1.client
        async with my_cache_1 as my_cache_2: # <<< new client created
            assert my_cache_1 is my_cache_2  # because of return self
            client_2 = my_cache_2
            assert client_1 != client_2
            assert my_cache_1.client == client_2
            # exit inner with, call __aexit__, calls await self.client.close(); i.e. client_2.close())
        assert my_cache_1.client != client_1 # it's still client_2
        # exit inner with, call __aexit__, calls await self.client.close(); i.e. client_2.close() again, which may error.
    # client_1.close() may never be called
```
I think that the `__aenter__` method should maybe handle checking if an existing connection is available and disposing of it if not, or else return a clone of `self` with a new client to ensure that the connection does get managed correctly.

pctablet505:
## What do these changes do?

`ValkeyCache.__aenter__` always created a brand new `GlideClient` and
overwrote `self.client`, even if the instance was already inside an
`async with` block. Nesting the same instance (e.g. `async with cache:
async with cache: ...`) silently replaced the connection, so the inner
block's exit closed the newer client while the outer block's exit then
closed it again (or the original connection leaked, never being closed).

`__aenter__` now raises `RuntimeError` if a client is already active,
and `__aexit__` clears `self.client` back to `None` after closing so
the same instance can still be used again sequentially (just not
nested).

Nesting the same `ValkeyCache` instance in overlapping `async with`
blocks now raises `RuntimeError` instead of silently leaking/double-closing
connections. Normal (non-nested) usage is unaffected.

## Related issue number

Fixes #1023

## Checklist


tests/ut/backends/test_valkey.py
Dreamsorcerer, line 1: Please use create_autospec(..., spec_set=True, instance=True) and patch(..., autospec=True, spec_set=True).

aiocache/backends/valkey.py
```diff
@@ -56,11 +59,19 @@ def __init__(
...
+                "ValkeyCache context is not reentrant. "
+                "Nesting the same instance in multiple 'async with' blocks "
+                "is not supported."
+            )
         self.client = await GlideClient.create(self.config)
         return self
 
     async def __aexit__(self, *args, **kwargs) -> None:
-        await self.client.close()
+        if self.client is not None:
```
Dreamsorcerer, line 72: This should be an assert, it should never be false.

